Issue #4, opened 2019-09-28
Download based on a geographic input?


Robinlovelace:
Thinking it may be useful for people using the package to download the smallest available file that exists for their region. That way the first argument to `get_geofabric()` (or similar) could be an sf object.

PR #11, opened 2019-10-08
Fix bugs in get_geofabric

agila5:
The first and third commit corrects a bug in `if(!continue)` since the output of `utils::menu` is not a logical value (i.e. TRUE or FALSE, the choices) but "The number corresponding to the selected item, or 0 if no choice was made." This means that we can substitute TRUE/FALSE with "Yes"/"No" and check if the user selectes "yes" (which corresponds to the value 1). I choose to modify the if clause to `if(continue != 1L)` since it now considers also the case in which the user entered 0 to `utils::menu` (i.e. no choice). 

The second and fourth commit correct a typo into the large_size check and a bug since we cannot put the large_size check inside the if (nrow(...) == 0) otherwise if it finds a match (e.g. get_geofabric("Africa") it doesn't check for large_size. 

I think you can check that now everthing works with this example: 

``` r
get_geofabric("HFIL")
```

### DIFF
--- a/R/get_geofabric.R
+++ b/R/get_geofabric.R
@@ -44,15 +44,16 @@ get_geofabric = function(
         stop("Search in geofabric_zones for a closer match.")
       }
     }
-    large_size = grepl(pattern = "G", x = geofabric_matches$size_pbf)
-    if(interactive() & ask & large_size) {
-      message("This is a large file ", geofabric_matches$size_pbf)
-      continue = utils::menu(choices = c("Yes", "No"), title = "Would you like to download this file?")
-      if(continue != 1L) {# same reasoning as before
-        stop("Aborted by user.")
-      }
-    }
     # add would you like to proceed message?
+  }
+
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Many thanks @agila5, sorry it took me so long to see this. I have other ideas for future fixes and now think this is worth publishing on CRAN, perhaps via rOpenSci. Thoughts?

agila5:
There is still an error in the PR 😭 

https://github.com/ITSLeeds/geofabric/blob/54b35d730eeb5593a9c314be3079b9c441bb199c/R/get_geofabric.R#L43

should be

``` r
# since the options are Yes/No, then Yes == 1L 
```
I don't know anything about CRAN submissions since I've never did it but I'm super happy for it. In my opinion, before the submission, we should add some tests and work on #4 (I have some simple ideas to solve it and we could discuss it later) plus any other idea you have.

Robinlovelace:
Doh, please fix with another PR, sorry I didn't spot that. Yes lots of ideas. `dl_geofabric()` seems key.

PR #21, opened 2019-11-12
Geofabric directory

Robinlovelace:


Robinlovelace:
Nudge @agila5 - seem good to you? On my desktop I've set the download directory so it always knows where to look before (repeatedly) downloading new data.

agila5:
I just arrived in ITS and I'm reading it now!

R/get_geofabric.R
```diff
@@ -5,6 +5,9 @@
 #' @param layer Character string telling `sf` which OSM layer to import.
 #' One of `points`, `lines` (the default), `multilinestrings`, `multipolygons` or `other_relations`
 #' @param download_directory Where to download the data? `tempdir()` by default.
+#' If you want to download your data into a persistend directory, set
```
agila5: I think you misspelled persistent. Moreover in my opinion something like: "If you want to download the Geofabrik data in a local and constant directory, set ..." is more clear.

R/get_geofabric.R
```diff
@@ -13,10 +16,9 @@
 #' @examples
 #' get_geofabric("isle of man")
 #' \donttest{
-#' get_geofabric("andorra")
-#' get_geofabric("west-yorkshire")
+#' get_geofabric(name = "andorra") # try other names, e.g. name = "west-yorkshire"
 #' # user asked to choose closest match when interactive
-#' get_geofabric("kdljfdl")
+#' get_geofabric("kdljfdl", ask = FALSE)
```
agila5: I don't understand why you set ask = FALSE if the example should show the possibility in interactive mode to check for closest match

Robinlovelace: Not a problem IMO:

```
get_geofabric("kdljfdl", ask = FALSE)
No exact matching geofabric zone. Best match is Poland (1.1 GB)
Nearest match to geofabric_zones$name is greater than threshold distance
NULL
```

PR #29, opened 2019-11-17
Fix #28

### LINKED ISSUE #28: Bug with get_geofabric and sfc data
Reprex: 
``` r
devtools::install_github("ITSLeeds/geofabric")
#> Skipping install of 'geofabric' from a github remote, the SHA1 (19a09498) has not changed since last install.
#>   Use `force = TRUE` to force installation

# packages
library(sf)
#> Linking to GEOS 3.6.1, GDAL 2.2.3, PROJ 4.9.3
library(geofabric)

# example
get_geofabric(
  name = st_sfc(st_point(c(9, 45)), crs = 4326)
)
#> Error in if (nrow(name) > 1) {: argument is of length zero
```

<sup>Created on 2019-11-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

The problem is here
https://github.com/ITSLeeds/geofabric/blob/19a094985f47d6dd8fb031a6df4bfc33e42b54ef/R/get_geofabric.R#L41-L42
since we cannot use `nrow` with `sfc` but only `length`. Should be easy fix.

agila5:
IMO the only problem is that we lose the possibility to consider additional information included in the sf object but I can't think of any useful example.

### DIFF
--- a/R/get_geofabric.R
+++ b/R/get_geofabric.R
@@ -39,7 +39,8 @@ get_geofabric = function(
   ) {
 
   if(inherits(name, "sf") | inherits(name, "sfc")) {
-    if(nrow(name) > 1) {
+    name <- sf::st_geometry(name)
+    if(length(name) > 1) {
       warning("Matching only based on the first feature.", immediate. = TRUE)
       message("Try sf::st_union() to convert into a single multi feature.")
     }
--- a/R/get_geofabric.R
+++ b/R/get_geofabric.R
@@ -39,7 +39,7 @@ get_geofabric = function(
   ) {
 
   if(inherits(name, "sf") | inherits(name, "sfc")) {
-    name <- sf::st_geometry(name)
+    name = sf::st_geometry(name)
     if(length(name) > 1) {
       warning("Matching only based on the first feature.", immediate. = TRUE)
       message("Try sf::st_union() to convert into a single multi feature.")


R/get_geofabric.R
```diff
@@ -39,7 +39,8 @@ get_geofabric = function(
   ) {
 
   if(inherits(name, "sf") | inherits(name, "sfc")) {
-    if(nrow(name) > 1) {
+    name <- sf::st_geometry(name)
```
Robinlovelace: Looks good to me, but pls change to 

```
name = sf::st_geometry(name)
```

for consistency.

agila5: Fixed!

agila5:
Anyway there is still something weird, i.e.:
``` r
# packages
library(sf)
#> Linking to GEOS 3.7.1, GDAL 2.4.2, PROJ 5.2.0
library(geofabric)

get_geofabric(
  st_sfc(
    st_point(c(9, 45)), # near Milan
    st_point(c(-1, 54)), # near Leeds
    crs = 4326
  )
)
#> Warning in get_geofabric(st_sfc(st_point(c(9, 45)), st_point(c(-1, 54)), :
#> Matching only based on the first feature.
#> Try sf::st_union() to convert into a single multi feature.
#> although coordinates are longitude/latitude, st_contains assumes that they are planar
#> The place is within these geofabrik zones: Europe, Great Britain, Italy, Britain and Ireland, England, Nord-Ovest, North Yorkshire
#> Selecting the smallest: North Yorkshire
#> Downloading http://download.geofabrik.de/europe/great-britain/england/north-yorkshire-latest.osm.pbf to 
#> /tmp/Rtmp2FibFb/North Yorkshire.osm.pbf
#> Old attributes: attributes=name,highway,waterway,aerialway,barrier,man_made
#> New attributes: attributes=name,highway,waterway,aerialway,barrier,man_made,maxspeed,oneway,building,surface,landuse,natural,start_date,wall,service,lanes,layer,tracktype,bridge,foot,bicycle,lit,railway,footway
#> Using ini file that can can be edited with file.edit(/tmp/Rtmp2FibFb/ini_new.ini)
```

<sup>Created on 2019-11-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

since the first feature should be matched with Milan, not Leeds. I'll check that in the afternoon. Should I create a new issue/PR or continue here?

PR #31, opened 2020-01-21
Keyvalue

Robinlovelace:


Robinlovelace:
Thanks for the review!

R/read_pbf.R
```diff
@@ -25,8 +38,16 @@ read_pbf = function(dsn,
     ini_new = make_ini_attributes(attributes = attributes, layer = layer, append = TRUE)
     writeLines(ini_new, ini_file)
   }
-  message("Using ini file that can can be edited with file.edit(", ini_file, ")")
   config_options = paste0("CONFIG_FILE=", ini_file)
+  if(!is.null(key)) {
+    if(is.null(value)) {
+      value = "*"
+    }
+    query = paste0("select ", key, " from ", layer, " where ", key, " = '", value,"'")
```
agila5: I'm sorry but I don't understand the query. Why do we use the same `key` parameter for select and where?

R/read_pbf.R
```diff
@@ -25,8 +38,16 @@ read_pbf = function(dsn,
...
+    res = sf::read_sf(dsn = dsn, layer = layer, options = config_options, query = query)
```
agila5: I don't understand this line. Why do we need a new sf::read_sf function call instead of adding the query parameter to the other sf::read_sf function?

R/read_pbf.R
Robinlovelace: Hmmm will take a look.

agila5: For example if `key` is equal to `"highway"`, `layer` is equal to `"lines"` and `"value"` is equal to `"motorway"` then `query` is equal to `"select highway from lines where highway = 'motorway'"` but I don't see why I should use the same variable for select and where.

Robinlovelace: The idea was to create dplyr-like filtering. However just telling people how to use the query parameters is a strong alternative (or perhaps parallel) strategy that keeps the user 'close to the metal' and avoids complication.

Robinlovelace: Good point @agila5 I see that `select *` is a better default. Will change. The point is to make it less words to type than: 

```
sf::read_sf(f, layer = "lines", query = "select * from lines where highway = 'cycleway'")
```

R/read_pbf.R
Robinlovelace: To allow `query = NULL`, which fails:

```
res_cycleways = sf::read_sf(f, layer = "lines", query = NULL)
Error in CPL_read_ogr(dsn, layer, query, as.character(options), quiet,  : 
  Not compatible with STRSXP: [type=NULL].
```

Robinlovelace:
Thanks for the post-merge comments @agila5. I've addressed them here: https://github.com/ITSLeeds/geofabric/commit/99adbe122ea74ad9c95f7a902ab9e44fd71e844d

The updated examples show why `read_pbf()` is useful: it reduces typing.

It's not perfect though. I was thinking of a lower level solution, a function called `gf_query()` that generates valid queries. Thoughts?

Issue #25, opened 2019-11-15
Bug when trying to download data from Central America


Robinlovelace:
Any ideas @agila5 ?

``` r
library(geofabric)
place_sf = sf::st_as_sf(data.frame(x = -86, y = 15, n = 1),
                        coords = c("x", "y"), crs = 4326)
osm_data_place = get_geofabric(place_sf)
#> although coordinates are longitude/latitude, st_contains assumes that they are planar
#> The place is within these geofabrik zones: Central America
#> Selecting the smallest:
#> Error in if (interactive() & ask & large_size) {: argument is of length zero
plot(osm_data_place["highway"])
#> Error in plot(osm_data_place["highway"]): object 'osm_data_place' not found
```

<sup>Created on 2019-11-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

agila5:
The problem is here 
https://github.com/ITSLeeds/geofabric/blob/170dda88c73d716d0d8fe38b09f6c3721d1d65b9/R/gf_find.R#L49
since in some cases geofabric_all_matches$level can be NA. Not sure why since the geofabric_zones.R file is quite messy and I don't fully understand that. 

I can try polish that first and then work on this issue.

Robinlovelace:
Good point. Overall we need to re-build the `geofabric_regions` data file to solve this at a higher level. Also, do you know a way to convert `MB` and `GB` file sizes into a single numeric unit (MB I suggest)?

agila5:
> Also, do you know a way to convert MB and GB file sizes into a single numeric unit (MB I suggest)?

Nope. It shouldn't be too difficult to write an adhoc function to translate something like `(3.4 GB)` into a numeric MB. I'll try 😅

agila5:
> Also, do you know a way to convert MB and GB file sizes into a single numeric unit (MB I suggest)?

I was thinking of something like this: 

``` r
# packages
library(tidyverse)

# example
size_pbf <- c("(3.4 GB)", "(29.0 MB)", "(7.4 GB)", "(709 MB)", "(365 MB)", "(20.5 GB)", "(8.7 GB)", "(1.7 GB)")

my_fix_size_pbf <- function(size_pbf) {
  GB_flag <- str_detect(size_pbf, "GB")
  size_as_test <- str_extract(size_pbf, "(?:(?=\\d+\\.)\\d+\\.\\d+|\\d+)")
[...]

PR #34, opened 2020-01-22
Change pkg name from geofabric to geofabrik. Fix #32

### LINKED ISSUE #32: Package name
geofabric or geofabrik https://www.geofabrik.de/data/download.html

agila5:
As the title says. The next step should be to change the git repo name, right?


README.Rmd
```diff
@@ -86,15 +86,16 @@ iow_lines_subset = sf::st_read(iow_gf_file, layer = "lines", query = query)
 plot(iow_lines_subset)
 ```
 
-You can also get your geofabric data based on a spatial query, for example in Honduras:
+<!-- TODO: FIX THE FOLLOWING PART SINCE IT DOESN'T WORK -->
```
agila5: The problem here is unrelated to this PR and is the same as #25

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
-Package: geofabric
+Package: geofabrik
 Type: Package
-Title: What the Package Does (Title Case)
```
Robinlovelace: Not quite CRAN-required Title Case but Will Change After Merging!

Robinlovelace:
Many thanks for a great set of changes @agila5 !

Issue #39, opened 2020-01-23
Weird result with query parameter


agila5:
I was working on the OGR_INTERLEAVED_READING problem (and I will submit a draft PR shortly) and I think I found something weird with the query parameter. I'm not sure if it's a bug or not and I can't easily reproduce that with geofabrik only but the following reprex should show the problem. 

``` r
library(sf)
#> Linking to GEOS 3.6.1, GDAL 2.2.3, PROJ 4.9.3

my_url <- "https://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf"
sf::st_read(my_url, layer = "points")
#> Reading layer `points' from data source `https://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf' using driver `OSM'
#> Warning in CPL_read_ogr(dsn, layer, query, as.character(options), quiet, :
#> GDAL Error 1: Too many features have accumulated in lines layer. Use
#> OGR_INTERLEAVED_READING=YES mode
#> Simple feature collection with 75666 features and 10 fields
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: -2.326617 ymin: 53.3142 xmax: -1.039338 ymax: 54.03102
#> epsg (SRID):    4326
#> proj4string:    +proj=longlat +datum=WGS84 +no_defs
sf::st_read(my_url, layer = "points", query = "select * from points")
#> Reading layer `points' from data source `https://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf' using driver `OSM'
#> Simple feature collection with 75666 features and 10 fields
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: -2.326617 ymin: 53.3142 xmax: -1.039338 ymax: 54.03102
#> epsg (SRID):    4326
#> proj4string:    +proj=longlat +datum=WGS84 +no_defs
```

<sup>Created on 2020-01-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

What's the difference between the two approaches? Why I don't see any warning if I add a query that shouldn't modify anything?

Robinlovelace:
Interesting. Perhaps the way that `sf` dispatches the call to GDAL when there is a query argument hides the warning? This seems like a potential `sf` issue. Thoughts @edzer (and sorry these OSM explorations are generating pesky questions!)?

edzer:
[...]

Issue #12, opened 2019-10-08
  GDAL Error 1: Too many features have accumulated in lines layer. Use OGR_INTERLEAVED_READING=YES mode


agila5:
Reprex of the error: 

``` r
library(geofabric)
get_geofabric(layer = "points")
#> No exact matching geofabric zone. Best match is West Yorkshire (28.7 MB)
#> Downloading http://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf to 
#> /tmp/RtmpfjodGC/west-yorkshire.osm.pbf
#> Old attributes: attributes=name,barrier,highway,ref,address,is_in,place,man_made
#> New attributes: attributes=name,barrier,highway,ref,address,is_in,place,man_made,building,natural,surface,source,power,amenity,shop,operator
#> Using ini file that can can be edited with file.edit(/tmp/RtmpfjodGC/ini_new.ini)
#> Warning in CPL_read_ogr(dsn, layer, query, as.character(options), quiet, :
#> GDAL Error 1: Too many features have accumulated in lines layer. Use
#> OGR_INTERLEAVED_READING=YES mode
```

<sup>Created on 2019-10-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Probably I'm missing something, but I'm not sure why it returns `Too many features have accumulated in lines layer` even if I set `layer = "points"`. Anyway I tried added the `OGR_INTERLEAVED_READING=YES` option to `st_read` but with no success: 

``` r
my_url <- "https://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf"
sf::st_read(my_url, layer = "points", options = "OGR_INTERLEAVED_READING=YES")
#> options:        OGR_INTERLEAVED_READING=YES 
#> Reading layer `points' from data source `https://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf' using driver `OSM'
#> Warning in CPL_read_ogr(dsn, layer, query, as.character(options), quiet, :
#> GDAL Error 1: Too many features have accumulated in lines layer. Use
#> OGR_INTERLEAVED_READING=YES mode
#> Simple feature collection with 72849 features and 10 fields
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: -2.326617 ymin: 53.3142 xmax: -1.039338 ymax: 54.03102
#> epsg (SRID):    4326
[...]

PR #40, opened 2020-01-23
Add gpkg approach for reading .osm.pbf files

agila5:
A few lines to show the "new" approach to read `.osm.pbf` files after the discussion in #12 In the next days (and as soon as I understand #39) I will write the missing examples, docs and so on.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -14,6 +14,7 @@ License: GPL-3
 Encoding: UTF-8
 LazyData: true
 RoxygenNote: 7.0.2
+Roxygen: list(markdown = TRUE)
 URL: https://itsleeds.github.io/geofabrik/, https://github.com/itsleeds/geofabrik
 BugReports: https://github.com/itsleeds/geofabrik/issues
 Depends: R (>= 3.4.0)
--- a/R/get_geofabrik.R
+++ b/R/get_geofabrik.R
@@ -1,4 +1,6 @@
-#' Download OSM data from [geofabrik.de](http://www.geofabrik.de/)
+#' Download OSM data from geofabrik
+#'
+#' Download OSM data from <http://download.geofabrik.de/>
 #'
 #' @inheritParams read_pbf
 #' @param name String or [`sf::sf`] spatial object of the geofabrik zone
--- a/R/read_pbf.R
+++ b/R/read_pbf.R
@@ -1,12 +1,12 @@
 #' Read pbf files with additional attributes
 #'
-#' Read pbf files (typically downloaded with [`get_geofabrik()`]) translating
-#' the `.osm.pbf`` file into a `.gpkg`` format. See details and the discussion
-#' in <https://github.com/OSGeo/gdal/issues/2100>
+#' Read pbf files (typically downloaded with [`get_geofabrik()`]) after
+#' translating the `.osm.pbf` format into a `.gpkg` format. See details.
 #'
 #' @inheritParams make_additional_attributes
 #' @param dsn The location of the file
-#' @param key Character string defining the key values to subset the data from, e.g. `"highway"`
+#' @param key Character string defining the key values to subset the data from,
+#'   e.g. `"highway"`.
 #' @param value The value(s) the `key` can take, e.g. `"cycleway"`
 #' @param selected_columns The columns to return in the output
 #' @param ini_file A modified version of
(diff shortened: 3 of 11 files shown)


Robinlovelace:
Looks good so far...

agila5:
IMO the new documentation is read now. What do you think? Please flag any typo or bad english grammar. 

I can't really show the differences between the previous and current approaches because of #39 (i.e. #12 gets fixed just by adding a query parameter to sf::read_sf as in the following reprex, I don't know why). 

``` r
library(sf)
#> Linking to GEOS 3.6.1, GDAL 2.2.3, PROJ 4.9.3

my_url <- "https://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf"
sf::st_layers(my_url, do_count = TRUE) #can't read multipolygons layer
#> Warning in CPL_get_layers(dsn, options, do_count): GDAL Error 1: Too many
#> features have accumulated in lines layer. Use OGR_INTERLEAVED_READING=YES mode
#> Driver: OSM 
#> Available layers:
#>         layer_name       geometry_type features fields
#> 1           points               Point    75915     10
#> 2            lines         Line String   174369      9
#> 3 multilinestrings   Multi Line String       46      4
#> 4    multipolygons       Multi Polygon       28     25
#> 5  other_relations Geometry Collection     7902      4
sf::st_read(my_url, layer = "multipolygons", query = "select * from multipolygons")
#> Reading layer `multipolygons' from data source `https://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf' using driver `OSM'
#> Simple feature collection with 443847 features and 25 fields
#> geometry type:  MULTIPOLYGON
#> dimension:      XY
#> bbox:           xmin: -2.189691 ymin: 53.4693 xmax: -1.180655 ymax: 53.97511
#> epsg (SRID):    4326
#> proj4string:    +proj=longlat +datum=WGS84 +no_defs
```

<sup>Created on 2020-01-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


I would still apply the .gpkg approach since it was suggested by the GDAL guys. 

I would also modify the following parts if you agree: 

1. The SQL query part should be included in the ogr2ogr conversion and that should make read reading function slightly faster since the query is performed before the conversion;
2. I would change sf::read_sf to sf::st_read or document the fact that get_geofabrik and read_pbf return tbl output.

Robinlovelace:
Hey @agila5 great work, looks good to me. Regarding your questions

> The SQL query part should be included in the ogr2ogr conversion and that should make read reading function slightly faster since the query is performed before the conversion;

Yes if it works that sounds like a good default. An argument like 'save_gpkg = FALSE` could be switched to make the `.pbf` files to `gpkg` files for future use, I was thinking, but only if that works. The read function should prefer to read in a `gpkg` file of the same name if there is one is my thinking.

Regarding 2) I suggest documenting the `tbl` output, it's more useful to have characters than factors for sure.

agila5:
> An argument like 'save_gpkg = FALSE` could be switched to make the `.pbf` files to `gpkg` files for future use, I was thinking, but only if that works. The read function should prefer to read in a `gpkg` file of the same name if there is one is my thinking.

I didn't thought about that but it makes sense. I would also set 'save_gpkg = TRUE`  and move the vectortranslate code to get_geofabrik function so that we save only the .gpkg file instead of the .osm.pbf file. I don't see the point in saving the same file in two different formats and I would choose to save only the gpkg file. Do you agree?

Robinlovelace:
I agree

Robinlovelace:
Great work by the way, the PR is looking strong!

agila5:
Yesterday I was thinking about the following idea, which is based on the previous comments. We divide the get_geofabrik function into three smaller function: the first one is used to download .osm.pbf data from geofabrik, the second one in used to translate the .osm.pbf data into .gpkg format and the third one is used to read the .gpkg format. IMO it's a reasonable approach and we could easily modify just the first function to include several providers (i.e. we should add a provider parameter, get_geofabrik becomes get_osm and so on) because of #32. 

Is that reasonable? If so I will update this PR as soon as possible. I'm super busy these days because I have to prepare a lot of materials for a Skype call at the end of the next week 😭

Robinlovelace:
Hey @agila5 you plan sounds reasonable. Modularity is good. I fully endorse the idea of breaking the function up into 3 smaller ones that can be joined together in `get_geofabric()`, an approach that seems to work well in the function `get_stats19()`.

Robinlovelace:
Hey @agila5 any update on this?

agila5:
As usual I forget about these things 😅 I think I will continue on the weekend, Saturday probably, because I think it will take a few hours.

PR #57, opened 2020-07-10
Providers vignette

Robinlovelace:


Robinlovelace:
This is good to go. Great news, your system works :tada: !

R/download.R
```diff
@@ -16,7 +16,7 @@
 oe_download <- function(
   file_url,
   file_basename = basename(file_url),
-  provider = infer_provider_from_url(file_url),
+  provider = "geofabrik",
```
agila5, line 19: The idea here was to "infer" the chosen provider from the input url. For example if the input is "whatever.geofabrik.de/iow.osm.pbf" then the provider will be "geofabrik"; if the input url is "whatever.bbbike.com/leeds.osm.pbf" then the provider is "bbbike" and the files are saved as "provider_filename.osm.pbf". Why it doesn't work?

R/download.R
```diff
@@ -41,13 +41,14 @@ oe_download <- function(
...
 
   if (!file.exists(file_path) || isTRUE(force_download)) {
-    if (interactive() && !is.na(file_size) && file_size >= max_file_size) {
-      message("This is a large file (", round(file_size / 1e+6), " MB)!")
-      continue <- utils::menu(
-        choices = c("Yes", "No"),
-        title = "Are you sure that you want to download it?"
-      )
-
+    if(provider == "geofabrik") {
```
agila5, line 44: I think that you can deduce the file_size also for bbbike but I will check it later

R/download.R
Robinlovelace: It caused an error message.

Robinlovelace: Try it. I agree with the usefulness of the function, but it crashed my code!

R/download.R
Robinlovelace: Possibly but I don't think we need size for bbbike because the biggest is a single city (London perhaps).

R/download.R
agila5: Okok, I will check it after merging the PR. The problem is that, at the moment, all the files are saved as "geofabrik_filename.osm.pbf" even if they are download from bbbike, right?

Robinlovelace: Ah possibly... I gets the data for sure though!

Issue #59, opened 2020-07-12
Document params


Robinlovelace:
Currently many of them just say `TODO`.

PR #60, opened 2020-07-12
Document verbose in oe_download for #59

Robinlovelace:


Robinlovelace:
Finally ready for review I think @agila5 more of a mission than I expected and has uncovered some other issues that we can fix in other commits/PRs.

agila5:
Thank you very much, tomorrow I will check it much better!

PR #72, opened 2020-07-14
Update geofabrik zones

Robinlovelace:


Robinlovelace:
Heads-up @agila5 the latest commit in this PR means that the code should work regardless of the `GDAL/PROJ` version the user has installed. Ready to merge I think but want you to check it before it goes into master.

agila5:
Great, thanks! So now we set the CRS in `load_provider_data()` and we can remove the extra code in `geofabrik_zones.R,` right?

Robinlovelace:
> So now we set the CRS in load_provider_data() and we can remove the extra code in geofabrik_zones.R, right?

Correct.

Issue #75, opened 2020-07-15
Uniform verbose parameters


agila5:
On a second thought I think that the idea of creating three different parameters (i.e. `oe_verbose`, `oe_quiet` and `quiet`) for showing a more detailed description of what's happening in the function is unnecessarily complicated. I would simply use `quiet` for all of them and, at the end, set `quiet = quiet` in `sf::st_read`. In this way you get all the extra messages or none of them.

Robinlovelace:
All or nothing sounds reasonable. Simplicity is good!

Issue #73, opened 2020-07-14
Test pbf download is time, memory and bandwidth consuming


Robinlovelace:
The Isle of Wight example we're using is only a few MB but is still slow compared with other examples.

I suggest creating a new zone called `test` in the `text_zones` object that points here: https://github.com/ITSLeeds/osmextractr/raw/master/inst/its-example.osm.pbf

That will speed-up our tests!

Sound like a plan @agila5?

Robinlovelace:
Seems to be fixed to me so closing :tada: 

```
its_leeds = oe_get("ITS Leeds", provider = "test", oe_verbose = TRUE)
The input place was matched with: ITS Leeds
trying URL 'https://github.com/ITSLeeds/osmextractr/raw/master/inst/its-example.osm.pbf'
Content type 'application/octet-stream' length 40792 bytes (39 KB)
==================================================
downloaded 39 KB

File downloaded!
Start with the vectortranslate operations on the input file!
Finished the vectortranslate operations on the input file!
Reading layer `lines' from data source `/tmp/RtmpXKYWRj/test_its-example.gpkg' using driver `GPKG'
Simple feature collection with 189 features and 9 fields
geometry type:  LINESTRING
dimension:      XY
bbox:           xmin: -1.562458 ymin: 53.80471 xmax: -1.548076 ymax: 53.81105
geographic CRS: WGS 84
```

PR #76, opened 2020-07-16
Add oe update

### LINKED ISSUE #64: oe_update()
I think a small function to download more recent versions of a file could be really useful. Building on https://github.com/ITSLeeds/osmextractr/issues/33 I think it could simply be:

``` r
library(osmextractr)
#> Data (c) OpenStreetMap contributors, ODbL 1.0. https://www.openstreetmap.org/copyright
oe_update = function(place, ...) {
  oe_get(place = place, ..., force_download = TRUE, download_only = TRUE)
}
oe_update("Isle of Wight")
#> [1] "/mnt/57982e2a-2874-4246-a6fe-115c199bc6bd/data/osm/geofabrik_isle-of-wight-latest.gpkg"
```

<sup>Created on 2020-07-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

agila5:
Fix #64. Show behaviour: 

``` r
library(osmextractr)
#> Data (c) OpenStreetMap contributors, ODbL 1.0. https://www.openstreetmap.org/copyright
oe_get("Portugal")
#> Reading layer `lines' from data source `/home/andrea/osmext_data/geofabrik_portugal-latest.gpkg' using driver `GPKG'
#> Simple feature collection with 1168329 features and 9 fields
#> geometry type:  LINESTRING
#> dimension:      XY
#> bbox:           xmin: -31.52917 ymin: 28.11984 xmax: -2 ymax: 51.28145
#> geographic CRS: WGS 84
oe_update(oe_verbose = TRUE)
#> This is a short description of some characteristics of the files stored in the download_directory: 
#>                                                                 size
#> /home/andrea/osmext_data/geofabrik_portugal-latest.gpkg    511664128
#> /home/andrea/osmext_data/geofabrik_portugal-latest.osm.pbf 222776445
#>                                                                          mtime
#> /home/andrea/osmext_data/geofabrik_portugal-latest.gpkg    2020-07-16 08:02:30
#> /home/andrea/osmext_data/geofabrik_portugal-latest.osm.pbf 2020-07-16 08:02:05
#>                                                                          ctime
#> /home/andrea/osmext_data/geofabrik_portugal-latest.gpkg    2020-07-16 08:02:30
#> /home/andrea/osmext_data/geofabrik_portugal-latest.osm.pbf 2020-07-16 08:02:05
#> 
#> Now the .osm.pbf files are going to be updated.
#> This is a short description of some characteristics of the updated  files stored in the download_directory: 
#>                                                                 size
#> /home/andrea/osmext_data/geofabrik_portugal-latest.osm.pbf 222776445
#>                                                                          mtime
#> /home/andrea/osmext_data/geofabrik_portugal-latest.osm.pbf 2020-07-16 08:02:59
#>                                                                          ctime
#> /home/andrea/osmext_data/geofabrik_portugal-latest.osm.pbf 2020-07-16 08:02:59
```

<sup>Created on 2020-07-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

There is no big difference before/after `oe_update` but you can see that `mtime `and `ctime `are different. There is still one big open point since I didn't implement the possibility to update the .gpkg files. Do you think that we need to add that functionality? The .gpkg files are the most important files here but they are quite difficult to be updated automatically since they could have been created with different layers and different options for the osmconf.ini file so I think that I could ruin them with an automatic update. 

Another question: What should the function return? Just TRUE if everything runs smoothly?


Robinlovelace:
Great work @agila5. In answer to your questions...

>  There is still one big open point since I didn't implement the possibility to update the .gpkg files. Do you think that we need to add that functionality?

Good point. We should reduce the probability of people accidentally reading-in old .gpkg files after running `oe_update()` to almost zero. Idea: add an argument `delete_gpkg` to the function that deletes all `.gpkg` files when `TRUE`, a sensible default I think. The code could look something like this:

```r
if(delete_gpkg) {
  gpkg_files = list.files(directory, pattern = paste0(file_name_base, "*.gpkg"))
  file.remove(gpkg_files)
  }
```

> Another question: What should the function return? Just TRUE if everything runs smoothly?

Or the path to the updated `.pbf` file may be better.

agila5:
Ok thanks, I will update the PR this evening

R/update.R
```diff
@@ -97,6 +111,8 @@ oe_update = function(
     )
     print(new_files_info[, c(1, 4, 5)])
   }
+
+  invisible(osmpbf_files)
```
Robinlovelace, line 115: Why invisible? I imagine it's good to see it printed but there may be unforseen disadvantages that I've not thought of.

agila5: Because I thought that it would not be interesting to see a bunch of paths unless you need to perform some operations on them (in which case you can save the result). It's reasonable?

agila5:
If you agree with the function and docs then I would merge. I will add the examples after #73 and #75

PR #89, opened 2020-07-21
Read pbf or gpkg with oe_get

### LINKED ISSUE #85: Read an existing file with oe_get()
I think it would be nice (and probably not that difficult) adding the possibility of reading/translating an existing `pbf`/`gpkg` file through `oe_get()`. Something like: 

``` r
oe_get("path_to_osm.pbf")
#> output
```

agila5:
Will fix #85.

### DIFF
--- a/R/get.R
+++ b/R/get.R
@@ -94,41 +94,65 @@ oe_get = function(
   skip_vectortranslate = FALSE,
   quiet = TRUE
 ) {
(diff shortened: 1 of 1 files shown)


agila5:
IMO it's ready for review. At the moment there is a lot of code duplication between oe_read and oe_get but that's for another PR (#90)

PR #93, opened 2020-07-28
Add oe_providers()

### LINKED ISSUE #86: Export oe_available_providers or something similar
Because I think it could be useful for printing the existing and available providers in some nice way (like a table with the colnames and so on)

agila5:
Fix #86. This is the result but I'm not 100% sure what should be displayed (apart for provider's names and database names): 

``` r
library(osmextract)
#> Data (c) OpenStreetMap contributors, ODbL 1.0. https://www.openstreetmap.org/copyright
oe_providers()
#> Check the corresponding help pages to read more details about the fields in each database
#>   available_providers   database_name number_of_zones number_of_fields
#> 1           geofabrik geofabrik_zones             430               14
#> 2              bbbike    bbbike_zones             235                9
```

<sup>Created on 2020-07-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Also the testthat::test is quite useless since it's the same thing as the example. Suggestions?


Robinlovelace:
Clearly a step in the right direction so merging, assuming you're not planning to add any more commits here.

agila5:
I forgot to merge 😅 Thanks!

PR #95, opened 2020-07-29
Update docs and tests for match.R

agila5:


agila5:
@Robinlovelace at the moment the tests check the same stuff than the examples but I think it's good enough for the moment and I can update the tests as soon as we sound the bugs

agila5:
@Robinlovelace that's all for match.R. I will never be able to review the other functions before tomorrow but I think a more realistic deadline for the submission is next friday.

PR #101, opened 2020-07-30
Intro vignette

Robinlovelace:


Robinlovelace:
Heads-up @agila5 this is a good first pass. I've left a couple of sections that I think you may be best placed to write because you led on `oe_read()` and the vectortranslate stuff but suggest that, if the tests pass, you do that on the master branch. Good to merge from my perspective pending your feedback.

Robinlovelace:
I prefer `oe_check_pattern()`. Simple = good!

Thanks @agila5, just saw your question.

agila5:
> I prefer oe_check_pattern(). Simple = good!

Ok, I will revert it back after the s2 checks

PR #102, opened 2020-07-31
use_coverage()

Robinlovelace:


agila5:
Ok, thanks! Do you want to merge now or wait for checking against s2?

Robinlovelace:
> Ok, thanks! Do you want to merge now or wait for checking against s2?

The aim is to add coverage results. Not sure if it will work. By merging we may find out so I say merge it and let's see :+1:

Issue #92, opened 2020-07-24
Logo


Robinlovelace:
Here are some ideas from my friend and housemate Pete:

![image](https://user-images.githubusercontent.com/1825120/88385330-075caa00-cda6-11ea-94fb-639e3f2b9e29.png)

agila5:
I didn't forget about the logo and I'd choose the top-left one. Maybe, if your roommate can help us again (btw: please thank him on my behalf) and as a hidden "bonus", I think it would be even better if the image depicted the WKB representation of Leeds' centroid instead of generic 0s and 1s: 

``` r
sf::st_as_binary(sf::st_geometry(osmextract:::oe_search("Leeds")))
#> [[1]]
#>  [1] 01 01 00 00 00 c2 1f 34 71 61 b3 f8 bf d9 5c 35 cf 11 e6 4a 40
#> 
#> attr(,"class")
#> [1] "WKB"
```

<sup>Created on 2021-02-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Hence, it could represent 01 01 00 00 00 in the first diagonal and so on. If Pete can't work on these ideas anymore, I'd simply choose the top-left image which is already really good!

Robinlovelace:
Will ask Pete...

Robinlovelace:
> I'd choose the top-left one.

I agree. No harm in doing a poll also...

Robinlovelace:
![image](https://user-images.githubusercontent.com/1825120/200123543-6c674448-1530-4ccc-821b-4242c13997dc.png)

Robinlovelace:
Final version as a .png:

![image](https://user-images.githubusercontent.com/1825120/200123605-1349adbe-41a0-4fd0-a23e-99b62b4f75ae.png)

PR #103, opened 2020-07-31
Dev sf ci

Robinlovelace:


Robinlovelace:
Just a test. I think it will be useful to test against the dev version of sf so heads-up @agila5

agila5:
Hi! Thank for your help. Just one question: this is just for testing or do you want to merge this PR at some point? 

btw: maybe it's easier to specify the dependency on the dev version of `sf` through [`usethis::use_dev_package()`](https://usethis.r-lib.org/reference/use_package.html)

Robinlovelace:
It's generating some interesting error messages:

https://github.com/ITSLeeds/osmextract/pull/103/checks?check_run_id=931672088#step:11:21

agila5:
I think the problem is somewhere between s2 and sf but I have no idea. `oe_match()` breaks when it's estimating which geofabrik areas are intersecting the input point (using EPSG: 4326), i.e. `provider[input_point, ]`.  At the moment I would just wait a couple of days more and leave the PR open as a remainder.

agila5:
Hi @Robinlovelace, sorry to bother you. I just checked the R CMD check of osmextract on windows with github version of sf and it's running without any warning or error (just unrelated notes) so I want to test again with this PR. How can I add here all the commits in the master branch?

Robinlovelace:
With the following commands in a system shell:

```
git checkout dev-sf-ci
git merge master
git push
```

Robinlovelace:
Good timing: I was just about to open an issue here!

agila5:
Hurray :tada: :tada: :tada:  Anyway, I wouldn't merge the PR since I think the double R CMD checks are unnecessarily "expensive". Maybe we can just leave it here and check the dev version of sf every now and then (like a few days before every major CRAN release). Or maybe we can merge it and change the .yaml file so that the extra checks are run onlyl if the text of the commit message include something like "test dev sf". What do you think?

Robinlovelace:
> Maybe we can just leave it here and check the dev version of sf every now and then (like a few days before every major CRAN release)

:+1:

agila5:
Just a random idea (that's probably an overkill for something as simple as the code here but it could be a useful test for other projects): maybe we could [schedule](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#scheduled-events) the dev sf ci workflow to run the first day of each month (or something similar). Thoughts?

agila5:
Hi! I think that we can close this PR since that the package is on CRAN now and we should be notified of any problem with dev-sf when the maintainers run the revdep checks, right?

Robinlovelace:
> Hi! I think that we can close this PR since that the package is on CRAN now and we should be notified of any problem with dev-sf when the maintainers run the revdep checks, right?

Not necessarily but I think we can close anyway. Great work getting it out there. Next up #92 + getting it out there!

PR #113, opened 2020-08-07
docs, examples and tests for oe_get

### LINKED ISSUE #90: Merge oe_read and oe_get
After #85. See https://github.com/ITSLeeds/osmextract/issues/85#issuecomment-662284987.

agila5:
Fix #90

### DIFF
--- a/R/get.R
+++ b/R/get.R
@@ -1,11 +1,16 @@
-#' Download, translate and read OSM extracts
+#' Download, translate and read OSM extracts from several providers
+#'
+#' This function is used to download, translate and read OSM extracts downloaded
+#' from several providers. It is a wrapper around `oe_match()` and `oe_read()`.
+#' Check the introductory vignette, the examples and the help pages of the
+#' wrapped functions to understand the details behind all parameters.
 #'
 #' @param place Description of the geographical area that should be matched to a
-#'   .osm.pbf file through the chosen `provider`. Can be either a length-1
+#'   `.osm.pbf` file through the chosen `provider`. Can be either a length-1
 #'   character vector, a length-1 `sfc_POINT` object or a numeric vector of
 #'   coordinates with length 2. In the latter case it is assumed that the EPSG
 #'   code is 4326, while you can use any EPSG code with an `sfc_POINT` object.
-#'   See details and examples.
+#'   See details and examples of `oe_match()`.
 #' @param layer Which `layer` should be read in? Typically `points`, `lines`
 #' (the default), `multilinestrings`, `multipolygons` or `other_relations`.
 #' @param provider Which provider should be used to download the data? Available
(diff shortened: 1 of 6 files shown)


R/get.R
```diff
@@ -1,11 +1,16 @@
-#' Download, translate and read OSM extracts
+#' Download, translate and read OSM extracts from several providers
+#'
+#' This function is used to download, translate and read OSM extracts downloaded
+#' from several providers. It is a wrapper around `oe_match()` and `oe_read()`.
+#' Check the introductory vignette, the examples and the help pages of the
+#' wrapped functions to understand the details behind all parameters.
```
Robinlovelace, line 6: :+1:

R/get.R
```diff
@@ -1,11 +1,16 @@
...
 #'
 #' @param place Description of the geographical area that should be matched to a
-#'   .osm.pbf file through the chosen `provider`. Can be either a length-1
+#'   `.osm.pbf` file through the chosen `provider`. Can be either a length-1
 #'   character vector, a length-1 `sfc_POINT` object or a numeric vector of
 #'   coordinates with length 2. In the latter case it is assumed that the EPSG
 #'   code is 4326, while you can use any EPSG code with an `sfc_POINT` object.
-#'   See details and examples.
+#'   See details and examples of `oe_match()`.
```
Robinlovelace, line 13: :+1: :+1:

R/get.R
```diff
@@ -60,32 +65,47 @@
...
-#' # Add another layer to the .gpkg file
-#' its_leeds_points = oe_get("ITS Leeds", provider = "test", layer = "points")
-#' summary(sf::st_geometry_type(its_leeds_points))
+#' # Download OSM extracts associated to a simple test.
+#' its = oe_get("ITS Leeds", provider = "test", quiet = FALSE)
+#' class(its)
+#' summary(sf::st_geometry_type(its))
+#'
+#' # Add another layer to the test file
+#' its_points = oe_get("ITS Leeds", provider = "test", layer = "points", quiet = FALSE)
```
Robinlovelace, line 78: :+1: :+1: :+1:

R/get.R
```diff
@@ -60,32 +65,47 @@
...
 #' oe_get("ITS Leeds", provider = "test", download_only = TRUE, skip_vectortranslate = TRUE)
-#' # add additional attributes
-#' im = oe_get("ITS Leeds", provider = "test", extra_attributes = "oneway")
+#'
+#' # Add additional attributes
+#' im = oe_get("ITS Leeds", provider = "test", extra_attributes = "oneway", quiet = FALSE)
 #' names(im)
+#'
+#' # The ITS Leeds file is stored on the github page of the package, so we
+#' # need to manually set the provider. This is not the case with other
```
Robinlovelace, line 90: Should we make it a standard provider? Probably not to avoid confusion.

R/get.R
```diff
@@ -60,32 +65,47 @@
...
+#' # standards providers.
 #' \dontrun{
+#' west_yorkshire = oe_get("West Yorkshire", quiet = FALSE)
+#' # If you run it again, the function will not download the file or convert it
+#' west_yorkshire = oe_get("West Yorkshire", quiet = FALSE)
+#' # match with coordinates (any EPSG)
+#' milan_duomo = sf::st_sfc(sf::st_point(c(1514924, 5034552)), crs = 3003)
+#' # Warning: the .pbf file is 400MB
+#' oe_get(milan_duomo, quiet = FALSE)
```
Robinlovelace, line 99: Got to get Milan in there!

R/read.R
```diff
@@ -67,17 +67,18 @@ oe_read = function(
   # If the input file_path is an existing .gpkg file is the easiest case since
   # we only need to read it:
   if (file.exists(file_path) && tools::file_ext(file_path) == "gpkg") {
+    # I need the following if to return the .gpkg file path in oe_get
+    if (isTRUE(download_only)) {
```
Robinlovelace, line 71: What's wrong with:

```r
     if (download_only) { 
```

?

R/get.R
agila5: I'm not sure I understand the question, sorry

R/read.R
agila5: Nothing. I used `isTRUE` and `isFALSE` because I read on their help page: 

```
isTRUE(x) is the same as { is.logical(x) && length(x) == 1 && !is.na(x) && x }; 
isFALSE() is defined analogously. 
Consequently, if(isTRUE(cond)) may be preferable to if(cond) because of NAs.
```

so the functions do not breaks if, for whatever reason, `download_only` is set equal to something that shouldn't be coercible to one logical value. Do you think it's a good idea?

Robinlovelace: Makes sense now, more robust code = good!

R/get.R
Robinlovelace: I recall that the test provider isn't in `oe_providers()` and wondered if it would be worth adding it but then thought maybe no.

agila5: It's not in `oe_providers()` since I explicitly excluded it: 

https://github.com/ITSLeeds/osmextract/blob/bc4ffd33a27bb961ea69579c193ca2a427df0571/R/providers.R#L51-L53

IMO `test` provider (which is just a link to the github ITS file) shouldn't be used outside of examples and tests, no?

Robinlovelace: > IMO test provider (which is just a link to the github ITS file) shouldn't be used outside of examples and tests, no?

Agreed.

Issue #100, opened 2020-07-30
Add support for ad-hoc areas 


agila5:
See https://protomaps.com/extracts

agila5:
See https://github.com/ITSLeeds/osmextract/pull/129#discussion_r497321084

PR #135, opened 2020-10-02
Add URL check in oe_read

### LINKED ISSUE #134: Check if input "looks like" a URL
The current approach in `oe_read` is to check if the input `file_path` points to an existing file, otherwise it assumes that it is a URL: 

https://github.com/ITSLeeds/osmextract/blob/1655ab4186c9a1402584234c3a9e808659940044/R/read.R#L69-L93

The problem, here, is that I think it is very easy to raise a (not-so-informative) error if the input `file_path` points to a non-existing file: 

``` r
library(osmextract)
#> Any product made from OpenStreetMap must cite OSM as the data source.
#> OpenStreetMap_fr data are taken from http://download.openstreetmap.fr/
oe_read("geofabrik_typo-in-path.osm.pbf")
#> Error in utils::download.file(url = file_url, destfile = file_path, mode = "wb", : scheme not supported in URL 'geofabrik_typo-in-path.osm.pbf'
```

<sup>Created on 2020-10-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

So I was thinking of adding a new `if` clause in `oe_read` to check if the input `file_path` "looks like" a URL (maybe with a [regex](https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url)). What do you think is the best approach here?

agila5:
Fix #134

### DIFF
--- a/R/read.R
+++ b/R/read.R
@@ -86,13 +86,9 @@ oe_read = function(
     # Add an if clause to check if file_path "looks like" a URL
     # See https://github.com/ITSLeeds/osmextract/issues/134 and
     # https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url
-    looks_like_url <- grepl(
-      pattern = "https?:\\/\\/(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-zA-Z0-9()]{1,6}\\b([-a-zA-Z0-9()@:%_\\+.~#?&//=]*)",
-      x = file_path,
-      perl = TRUE
-      )
+    like_url <- is_like_url(file_path)
 
-    if (!looks_like_url) {
+    if (!like_url) {
       stop(
         "The input file_path does not correspond to any existing file ",
         "and it doesn't look like a URL.",
--- a/R/read.R
+++ b/R/read.R
@@ -86,7 +86,7 @@ oe_read = function(
     # Add an if clause to check if file_path "looks like" a URL
     # See https://github.com/ITSLeeds/osmextract/issues/134 and
     # https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url
-    like_url <- is_like_url(file_path)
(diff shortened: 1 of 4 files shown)


R/read.R
```diff
@@ -81,6 +82,25 @@ oe_read = function(
...
   if (!file.exists(file_path)) {
+
+    # Add an if clause to check if file_path "looks like" a URL
+    # See https://github.com/ITSLeeds/osmextract/issues/134 and
+    # https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url
+    looks_like_url <- grepl(
+      pattern = "https?:\\/\\/(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-zA-Z0-9()]{1,6}\\b([-a-zA-Z0-9()@:%_\\+.~#?&//=]*)",
+      x = file_path,
+      perl = TRUE
+      )
```
Robinlovelace, line 93: Worth abstracting this into a separate unexported function, so that these lines could be replaced by this (for example)?:

```r
     looks_like_url = is_url(file_path)
```

Robinlovelace (changes requested):
Minor suggested change to create a new un-exported function that would be easier to use elsewhere.

R/read.R
agila5: 👍

agila5: Done!

Robinlovelace (approved):

Robinlovelace (approved):

Robinlovelace:
Great work @agila5 that should make it easier to take ad-hoc extracts for #100. Minor note, I replaced `<-` with `=` for consistency.

agila5:
> Great work @agila5 that should make it easier to take ad-hoc extracts for #100. Minor note, I replaced <- with = for consistency.

I will always forget about that 😅 

Self note: Check the R code in https://github.com/jimhester/urlchecker if, for whatever reason, the code in that PR doesn't work

PR #136, opened 2020-10-21
Add checks for ... and st_read

agila5:
Previously: 

``` r
remotes::install_github("ITSLeeds/osmextract")
library(osmextract)
oe_get("Isle of Wight", stringAsFactor = TRUE)
#> Error in st_sf(x, ..., agr = agr, sf_column_name = sf_column_name): no simple features geometry column present
```

<sup>Created on 2020-10-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Now: 

``` r
remotes::install_github("ITSLeeds/osmextract", "test_ellipsis")
library(osmextract)
oe_get("Isle of Wight", stringAsFactor = TRUE)
#> Warning: The following arguments are probably misspelled: stringAsFactor
#> Error in st_sf(x, ..., agr = agr, sf_column_name = sf_column_name): no simple features geometry column present
```

<sup>Created on 2020-10-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I'm not a big fan of the current solution since it's (probably) unnecessarily complicated but I couldn't think of any simpler approach.

### DIFF
--- a/R/read.R
+++ b/R/read.R
@@ -83,14 +83,14 @@ oe_read = function(
     # use names(formals("st_read.character", envir = getNamespace("sf")))
     any(
       names(list(...)) %!in%
-      names(formals("st_read.character", envir = getNamespace("sf")))
+      names(formals(get("st_read.character", envir = getNamespace("sf"))))
     )
   ) {
     warning(
       "The following arguments are probably misspelled: ",
       setdiff(
         names(list(...)),
-        names(formals("st_read.character", envir = getNamespace("sf")))
+        names(formals(get("st_read.character", envir = getNamespace("sf"))))
       ),
       call. = FALSE,
       immediate. = TRUE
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Good defensive programming @agila5, happy with the approach, let's see if it passes (I guess it will)!

agila5:
> Good defensive programming @agila5, happy with the approach, let's see if it passes (I guess it will)!
Thanks! 

It failed because the `envir` argument was added to `formals()` in R 3.6.0 😅 I will fix that in 10 minutes

Robinlovelace:
> It failed because the `envir` argument was added to `formals()` in R 3.6.0 sweat_smile I will fix that in 10 minutes

Or we could just depend on a later version of R.

agila5:
This is just slightly more complicated than the previous approach and works in 3.5.0 🎉

Robinlovelace:
:tada: :tada:

Issue #143, opened 2020-11-03
Download timeout too low


dabreegster:
Hi! I'm getting a crash when running `osmextract::oe_get(place = "west yorkshire")`:

```
Error in utils::download.file(url = file_url, destfile = file_path, mode = "wb",  : 
  download from 'https://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf' failed
Calls: <Anonymous> -> oe_read -> oe_download -> <Anonymous>
In addition: Warning messages:
1: st_crs<- : replacing crs does not reproject data; use st_transform for that 
2: In utils::download.file(url = file_url, destfile = file_path, mode = "wb",  :
  downloaded length 14860288 != reported length 33826429
3: In utils::download.file(url = file_url, destfile = file_path, mode = "wb",  :
  URL 'https://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf': Timeout of 60 seconds was reached
```

In the 60s timeout, my connection only grabbed about 15MB of the file. I'm new to R and am having trouble figuring out where `utils::download.file` is actually defined. Where's the appropriate place to add a timeout? Also, could the underlying HTTP fetcher possibly be changed to one that includes a progress bar and shows the total file size, to reassure users on slower connections?

Thanks!

agila5:
Hi @dabreegster! 

Could you try running the following code:  `osmextract::oe_get("west yorkshire", quiet = FALSE, force_download = TRUE)` ? The argument `quiet = FALSE` should print some (hopefully) informative messages, while the parameter `force_download = TRUE` will force the (re)download of the `.osm.pbf` file in case something wrong happened during an old download process. 

This is the output on my laptop: 

``` r
osmextract::oe_get("west yorkshire", quiet = FALSE, force_download = TRUE)
#> The input place was matched with: West Yorkshire
#> File downloaded!
#> The corresponding gpkg file was already detected. Skip vectortranslate operations.
#> Reading layer `lines' from data source `C:\Users\Utente\Documents\osm_data\geofabrik_west-yorkshire-latest.gpkg' using driver `GPKG'
#> Simple feature collection with 3860 features and 9 fields
#> geometry type:  MULTILINESTRING
#> dimension:      XY
#> bbox:           xmin: -1.786844 ymin: 53.70248 xmax: -1.291864 ymax: 53.94313
#> geographic CRS: WGS 84
```

[...]

PR #145, opened 2020-11-04
Describe how to workaround downloads timing out. #143

dabreegster:
Is there github automation that regenerates `README.md`, or do I run something?


agila5:
Thanks for the PR. 

> Is there github automation that regenerates README.md, or do I run something?

If you are using RStudio you just need to click the "knit" button in the "Source" panel or press "Ctrl + Shift + K", otherwise you need to run `rmarkdown::render("README.Rmd")`. More details [here](https://rmarkdown.rstudio.com/authoring_quick_tour.html).

dabreegster:
I regenerated the readme, but the diff was quite large, including some of the figures. Expected?

README.md
```diff
@@ -114,12 +72,11 @@ Load the package with:
...
 ```
 
-To use alongside functionality in the `sf` package, we also recommend
-attaching this geographic data package as follows:
+To use alongside functionality in the `sf` package, we also recommend attaching this geographic data package as follows:
 
 ``` r
 library(sf)
-#> Linking to GEOS 3.8.0, GDAL 3.0.4, PROJ 6.3.1
+#> Linking to GEOS 3.6.2, GDAL 2.2.3, PROJ 4.9.3
```
Robinlovelace, line 79: Surprised it works with such old versions of GDAL/PROJ, good to know. FYI @dabreegster there is info here on getting more up-to-date OSGeo packages on Linux (not sure what OS you're running): https://github.com/r-spatial/sf#linux

Massive thank you for the PR btw!

agila5:
> I regenerated the readme, but the diff was quite large, including some of the figures. Expected?

More or less yes, since the images are generated using a `.osm.pbf` file of the Isle of Wight that is updated daily, so the images will change (slight) if you run the README on two different days. See [here](http://download.geofabrik.de/europe/great-britain/england/isle-of-wight.html). The other differences are simply related to the formatting of the `.md` file (i.e. several sentences were collapsed on the same line), probably caused by different versions of `knitr`/`pandoc`. That shouldn't be relevant.  

Thanks! Feel free to open a new issue if you find another bug or if you have any other question/doubt.

README.md
dabreegster: Ah, thanks for the pointers! I'm on Ubuntu 18.04 and have hit issues with old gdal before, but hadn't found the more up-to-date PPA.

PR #148, opened 2020-11-05
Update CI - faster runs, simpler set-up

Robinlovelace:


agila5:
Thanks! I was thinking more or less the same a few days ago. I don't understand most of the changes in the .yaml file but it works 🎉

.github/workflows/R-CMD-check.yaml
```diff
@@ -1,16 +1,19 @@
...
   pull_request:
     branches:
+      - main
       - master
 
 name: R-CMD-check
 
 jobs:
   R-CMD-check:
-    if: "!(contains(github.event.head_commit.message, 'skip ci')||contains(github.event.head_commit.message, 'ci skip')||contains(github.event.head_commit.message, 'skip-ci')||contains(github.event.head_commit.message, 'ci-skip')||contains(github.event.head_commit.message, 'skip_ci')||contains(github.event.head_commit.message, 'ci_skip'))"
```
agila5, line 13: Just one question. Would you mind "restoring" this line? I'm not 100% sure that it still works ( will test it later) but I used that to skip the R CMD checks for the simple commits (change the docs and so on)

agila5 (approved):

.github/workflows/R-CMD-check.yaml
Robinlovelace: Sure, will do.

Robinlovelace:
Read to go I think, if it passes! :crossed_fingers:

Issue #97, opened 2020-07-29
Add support for new classes in oe_get


agila5:
After #35 

- [x] sf
- [x] sfc_POLYGON
- [x] sfc_LINESTRING
- [ ] sp objects

agila5:
See also https://github.com/ITSLeeds/osmextract/pull/129#discussion_r497319224 for spatial matching using something different from a `sfc_POINT`

agila5:
I plan to work on `sfc_POLYGON` and `sfc_LINESTRING` extensions during the weekend (for a new project). The main question here is: how to determine the matched geographical area when we are comparing a POLYGON with a LINESTRING or a POLYGON with a POLYGON? The problem is that a POLYGON or a LINESTRING could intersect two or more geographical areas determined by a provider. 

The simplest solution should be "greatest overlapping area + a warning message", any other suggestion?

Robinlovelace:
Good thinking @agila5. I'm confident that this code will handle all `sf`/`sfc` classes, finding all zones that *completely contain* the geographic object, regardless of whether that's a line, polygon or point cloud:

https://github.com/ITSLeeds/osmextract/blob/3b927dd66f006c00d63e12424614696af1fede63/R/gf_find.R#L35-L50

The code should automatically select the smallest area that matches to avoid it downloading the world after every geographic query!

agila5:
Thanks, I will start from that code!

Robinlovelace:
I'm happy to give this a go...

Robinlovelace:
For now the following works : )

```
osm_luton = osmextract::oe_get(sf::st_centroid(region_luton))
```

Robinlovelace:
Would be great for this to work for polygons also, just tried this:

```
buildings = osmextract::oe_get(study_area, layer = "multipolygons")
Error: At the moment there is no support for matching objects of class sf. Feel free to open a new issue at github.com/itsleeds/osmextract
```

agila5:
> Would be great for this to work for polygons also, just tried this:

The easy solution would be to take the centroid and then fall back to the sfc_POINT method, is that reasonable?

Robinlovelace:
> The easy solution would be to take the centroid and then fall back to the sfc_POINT method, is that reasonable?

I think even simpler than that would be an `st_contains()` that would work with any geometry type and also ensure that larger regions are selected, e.g. for roads that cross 2 regions. I'm up for giving it a bash.

Robinlovelace:
[...]

PR #165, opened 2021-01-16
Add level argument

agila5:
I think this is the last step before writing the answer to rOpenSci reviews.


R/get.R
```diff
@@ -39,6 +39,10 @@
 #'   [oe_match()] to understand why this parameter is important. Ignored if
 #'   `place` is not a character vector since the matching is performed through a
 #'   spatial operation.
+#' @param level The desired hierarchical level in case of spatial matching. We
```
Robinlovelace, line 42: I suggest being more explicit, e.g.:

```
#' @param level An integer representing the desired hierarchical level in case of spatial matching.
#'    For the `geofabrik` provider, for example, `1` corresponds with continent-level datasets, `2` countries, `3` corresponds to regions.
```

Specific + examples = good here IMO.

Robinlovelace (commented):
Looks good. I would like to fix #97 before review, I can give that a go now if you'd like @agila5 ?

agila5:
Ok, I will start working on that issue, no worries!

R/get.R
agila5: 👍

agila5:
Hi @Robinlovelace. The approach is more or less the same as what we discussed yesterday. I added `st_combine` since the previous version returned the wrong result in case of multiple `sfg` objects. 

Current output: 
``` r
# packages
library(sf)
#> Linking to GEOS 3.8.0, GDAL 3.0.4, PROJ 6.3.1
library(osmextract)
#> Check the package website, itsleeds.github.io/osmextract, for more details.

# a random LINESTRING in Leeds
linestring_chapeltown <- st_sfc(
  st_linestring(
    rbind(c(-1.549077, 53.800755), c(-1.560897, 53.810830))
  ), 
  crs = 4326
)
suppressWarnings(oe_match(linestring_chapeltown))
#> although coordinates are longitude/latitude, st_contains assumes that they are planar
#> The input place was matched with multiple geographical areas.
#> Selecting the smallest administrative unit. Check ?oe_match for more details.
#> The input place was matched with multiple geographical areas. Selecting the area whose centroid is closest to the input place.
#> although coordinates are longitude/latitude, st_nearest_points assumes that they are planar
#> $url
#> [1] "https://download.geofabrik.de/europe/great-britain/england/west-yorkshire-latest.osm.pbf"
#> 
#> $file_size
#> [1] 32270118

# uk -> Italy
linestring_uk_italy <- st_sfc(
  st_linestring(
    rbind(c(-1.549077, 53.800755), c(11, 45))
  ), 
  crs = 4326
)
suppressWarnings(oe_match(linestring_uk_italy))
#> although coordinates are longitude/latitude, st_contains assumes that they are planar
#> $url
#> [1] "https://download.geofabrik.de/europe-latest.osm.pbf"
#> 
#> $file_size
#> [1] 23245241740

# lnd in spData
suppressWarnings(oe_match(spData::lnd))
#> although coordinates are longitude/latitude, st_contains assumes that they are planar
#> The input place was matched with multiple geographical areas.
#> Selecting the smallest administrative unit. Check ?oe_match for more details.
#> The input place was matched with multiple geographical areas. Selecting the area whose centroid is closest to the input place.
#> although coordinates are longitude/latitude, st_nearest_points assumes that they are planar
#> $url
#> [1] "https://download.geofabrik.de/europe/great-britain/england/greater-london-latest.osm.pbf"
#> 
#> $file_size
#> [1] 59746355
```

<sup>Created on 2021-01-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

If you agree, I will update the docs, merge and prepare the response for rOpenSci.

Robinlovelace (approved):
Great work @agila5, very happy with this.

Robinlovelace:
(Although should we suppress warnings inside the function?)

Robinlovelace:
In any case, this is good to merge :tada:

agila5:
> (Although should we suppress warnings inside the function?)

I thought about that and IMO no because the warnings should disappear as soon as they finish adopting S2 in sf. Nevertheless, the warnings are not that relevant in this particular case and could be silenced. What do you think?

Robinlovelace:
> What do you think?

I suspect non S2 will be around for a while but either work for me.

PR #180, opened 2021-03-18
Test actions

### LINKED ISSUE #4: Download based on a geographic input?

Robinlovelace:


agila5:
FYI: the GHA should work now.

PR #192, opened 2021-04-11
Add gdal tests

Robinlovelace:


agila5:
In the end, I implemented this step using a regex (#202) so I think we can close here.

Robinlovelace:
:+1:

Issue #178, opened 2021-02-22
Add clip argument to oe_get()


Robinlovelace:
The example here shows the functionality: https://docs.ropensci.org/osmextract/articles/osmextract.html#vectortranslate_options

```r
la_valletta = st_sfc(st_point(c(456113.1, 3972853)), crs = 32633) %>%
  st_buffer(5000)

# 2. Define the vectortranslate options
my_vectortranslate = c(
  "-f", "GPKG", 
  "-overwrite", 
  "-t_srs", "EPSG:32633",
  "-select", "highway",
  "-where", "highway IN ('primary', 'secondary', 'tertiary', 'unclassified')",
  "-clipdst", st_as_text(la_valletta),
  "-nlt", "PROMOTE_TO_MULTI",
  "lines"
)
```

I think it would be good to expose a `clipdst` argument, that could take an sf object directly. Thoughts @agila5 ?

Use case: getting data for specific cities in a country like India that seems to have few (if any) subdivisions, meaning reading in GB of data without a spatial filter.

agila5:
Hi @Robinlovelace. Yes, I like the idea, which is also linked with #142. I'm a little busy these days (fixing the last parts of PhD dissertation for the defence) but I will tackle both issues as soon as possible.

agila5:
One question. There are several arguments in `ogr2ogr` that could be useful during the vectortranslate process (e.g. `-where`, `-clipsrc`, `-clidst`, `-select`, `-sql`, `-select` ecc). I don't like the idea of creating an ad-hoc argument in `oe_vectortranslate()` for each `ogr2ogr` argument, since that may be really confusing. So I think that the arguments for `ogr2ogr` could be specified using a named list. For example: 

``` r
oe_get("somewhere", vectortranslate_options = list(where = "some query", clipsrc = some-polygon))
```

The default value for `vectortranslate_options` is `list()`, which implies that the options are automatically set by the function (i.e. the current behaviour). Opinions?

Robinlovelace:
A named list sounds sensible to me.

luukvdmeer:
What about arguments without a value, like `--overwrite`? Maybe makes sense to use boolean values for those, like in `list(overwrite = TRUE)`? And for the 'Open options' and 'Layer creation options' use nested lists?

I think it is clearer than the current structure with `c()`. And also much easier to check if a certain option is in there or not.

Robinlovelace:
Sounds like a new function would be good, e.g.:

```r
[...]

PR #197, opened 2021-05-05
Starter for 10 on #178

Robinlovelace:
See https://github.com/ropensci/osmextract/issues/178


Robinlovelace:
Heads-up @agila5. Any initial thoughts on the approach welcome.

PR #213, opened 2021-06-23
Add useR2021 slides

agila5:
The slides are also added to the `.Rbuildignore` so I don't need to add any extra dep (also they are not really rendered as a vignette ATM)


agila5:
<details>
Immagini: 
raw_iow
small-iow
clip-iow
spat-iow
pred_severe-eps-converted-to-1
pred-spat-int-1-1
</details>

PR #218, opened 2021-07-06
Add oe_get_network

agila5:
Just the first ideas. Will need to fix the filters + several intermediate checks. 

@Robinlovelace what do you think about the name of the function?


Robinlovelace:
It sounds good. An alternative would be `oe_get_network()` - have a very mild preference for that but would be interested to see what others think. Great to see use of the querying to get that data quickly!

agila5:
@Robinlovelace can you check the options set for the three modes of transport? If those are ok, I just need to document the function.

R/get-transport.R
```diff
@@ -0,0 +1,84 @@
...
+    AND
+    (access IS NULL OR access NOT IN ('private', 'no'))
+    AND
+    (highway IS NOT NULL AND
+    highway NOT IN (
+    'abandonded', 'bus_guideway', 'construction', 'cycleway',
+    'planned', 'platform', 'proposed', 'raceway'
+    ))
+    AND
+    (foot IS NULL OR foot != 'no')
```
Robinlovelace: I was going to sat that motorways cannot be cycled on with reference to https://wiki.openstreetmap.org/wiki/Tag:highway%3Dmotorway 

But then I read there that `highway=motorway` implies that `bicycle=no`.

Do we no if that tag gets auto tagged?

Robinlovelace: Looking at a motorway in Leeds that you definitely cannot cycle on I don't think it does get auto added: https://www.openstreetmap.org/way/400349319

R/get-network.R
```diff
@@ -0,0 +1,212 @@
...
+    layer = "lines",
+    extra_tags = c("access", "bicycle", "service"),
+    vectortranslate_options = c(
+    "-where", "
+    ((highway IS NOT NULL AND highway NOT IN (
+    'abandonded', 'bus_guideway', 'byway', 'construction', 'corridor', 'elevator',
+    'fixme', 'escalator', 'gallop', 'historic', 'no', 'planned', 'platform',
+    'proposed', 'raceway', 'steps'
+    )) OR (
+    highway IN ('motorway', 'motorway_link', 'footway',
```
Robinlovelace, line 81: OK I see motorway is excluded, great.

Robinlovelace (approved):
This is a great start.

Minor comments:

- [ ] In terms of documentation it's worth mentioning and adding a link to the source for the filters
- [ ] One for later: document how to create a custom network creation function, e.g. `net_wheelchair()`
- [ ] Mixed styles, the package has `=` and `<-` assignment on this branch

These are minor points. Overall big :+1:

agila5:
Thanks for the comments! Tomorrow I will fix the docs.

.github/workflows/R-CMD-check.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
       matrix:
         config:
           - {os: windows-latest, r: 'release'}
-          - {os: macOS-latest, r: 'release'}
+          # - {os: macOS-latest, r: 'release'}
```
Robinlovelace, line 28: FYI this can be re-added if you add the lines shown here to the .yaml file: 

https://github.com/a-b-street/abstr/commit/a37447737b8abf68301699ae2bfcdafb45587545

I suggest this is a separate issue, just mentioning here as an FYI.

R/get-network.R
```diff
@@ -0,0 +1,270 @@
...
+#' @param mode A character string denoting the desired mode of transport. Can be
+#'   abbreviated.
+#' @param ... Additional arguments passed to `oe_get()` such as `boundary` or
+#'   `force_download`.
+#'
+#' @return An `sf` object.
+#' @export
+#'
+#' @details The `cycling` mode of transport (i.e. the default value for `mode`
+#'   parameter) selects the OSM ways that meet the following conditions:
```
Robinlovelace, line 22: Great to have these details.

R/get-network.R
```diff
@@ -0,0 +1,275 @@
...
+#' @inheritParams oe_get
+#' @param mode A character string denoting the desired mode of transport. Can be
+#'   abbreviated. Currently cycling (the default), driving and walking are supported.
+#' @param ... Additional arguments passed to `oe_get()` such as `boundary` or
+#'   `force_download`.
+#'
+#' @return An `sf` object.
+#' @export
+#'
+#' @details The definition of usable transport network was taken from
```
Robinlovelace, line 16: Is this correct @agila5 ? If we modify the rules we could say 'was inspired by' but as far as I'm aware this is currently like-for-like.

R/get-network.R
```diff
@@ -0,0 +1,275 @@
...
+#'   https://github.com/ropensci/osmextract/issues/153
+#'
+#'   Feel free to start a new issue in the [github
+#'   repo](https://github.com/ropensci/osmextract) if you want to suggest
+#'   modifications to the current filters or propose new values for alternative
+#'   modes of transport.
+#'
+#' @seealso [`oe_get()`]
+#'
+#' @examples
```
Robinlovelace, line 78: Great examples.

R/get-network.R
```diff
@@ -0,0 +1,275 @@
...
+    walking = load_options_walking(place),
+    driving = load_options_driving(place)
+  )
+
+  # Check the other arguments supplied by the user
+  dots_args = list(...)
+  oe_get_options = check_args_network(dots_args, oe_get_options)
+
+  # Run oe_get
+  do.call(oe_get, oe_get_options)
```
Robinlovelace, line 107: Clever! I learned something from this, I think, all the named args are passed here, right?

R/get-network.R
```diff
@@ -0,0 +1,275 @@
...
+}
+
+# The following functions are used to load several ad-hoc vectortranslate
+# options according to a specific mode of transport. The choices are documented
+# in oe_get_network() and are based on the following documents:
```
Robinlovelace, line 112: Big :+1: for including these links, may make my previous comment about osmnx redundant.

Robinlovelace (approved):
Overall great PR, very useful new functionality and ready to ship now :ship: 

My comments are minor and can be tidied up post merge - feel free to change any of my commits to this branch also @agila5. Thanks for great work on this :tada:

.github/workflows/R-CMD-check.yaml
agila5: 👍

R/get-network.R
agila5: It was not exactly "taken" from osmnx but I will slightly change that in master branch, 👍

R/get-network.R
agila5: Yup! Thanks also to @luukvdmeer and @loreabad6, I learned that when reading the code behind `plot.sfnetwork` 😅

PR #223, opened 2021-08-02
Set timeout in oe_download

### LINKED ISSUE #222: `oe_download()` could internally set a `timeout` value
From the help page of `download.file()`: 

> The timeout for many parts of the transfer can be set by the option timeout which defaults to 60 seconds. This is often insufficient for downloads of large files (50MB or more) and so should be increased when download.file is used in packages to do so. Note that the user can set the default timeout by the environment variable R_DEFAULT_INTERNET_TIMEOUT in recent versions of R, so to ensure that this is not decreased packages should use something like

    options(timeout = max(300, getOption("timeout")))
  
> (It is unrealistic to require download times of less than 1s/MB.)

Most of the .pbf files are larger than 50MB. See also the discussion in #143

agila5:
Fix #222.

### DIFF
--- a/R/download.R
+++ b/R/download.R
@@ -133,6 +133,9 @@ oe_download = function(
       stop("Aborted by user.")
     }
 
+    old_opts <- options(timeout = max(300, getOption("timeout")))
+    on.exit(options(old_opts))
+
     utils::download.file(
       url = file_url,
       destfile = file_path,
--- a/R/download.R
+++ b/R/download.R
@@ -133,7 +133,7 @@ oe_download = function(
       stop("Aborted by user.")
     }
 
-    old_opts <- options(timeout = max(300, getOption("timeout")))
+    old_opts = options(timeout = max(300, getOption("timeout")))
     on.exit(options(old_opts))
 
     utils::download.file(
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # osmextract (development version)
 
+### MINOR CHANGES
+
+* Following the suggestion in `?download.file`, the function `oe_download()` automatically sets a `timeout` greater than 60 seconds. The original value is reset when `oe_download()` function exits. 
+
 # osmextract 0.3.1
 
 ### MAJOR CHANGES
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ### MINOR CHANGES
 
-* Following the suggestion in `?download.file`, the function `oe_download()` automatically sets a `timeout` greater than 60 seconds. The original value is reset when `oe_download()` function exits. 
+* Following the suggestion in `?download.file`, the function `oe_download()` automatically sets a `timeout` greater than 60 seconds. The original value is reset when `oe_download()` function exits (#222)
 
 # osmextract 0.3.1
 
--- a/README.Rmd
+++ b/README.Rmd
@@ -191,14 +191,16 @@ You can always check the default `download_directory` used by `oe_get()` with:
 oe_download_directory()
 ```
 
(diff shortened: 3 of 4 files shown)


R/download.R
```diff
@@ -133,6 +133,9 @@ oe_download = function(
       stop("Aborted by user.")
     }
 
+    old_opts <- options(timeout = max(300, getOption("timeout")))
```
Robinlovelace: Looks good, just tested and I get:

```
> res = options(timeout = max(300, getOption("timeout")))
> res
$timeout
[1] 60
```

Idea: add an argument `timeout` so the user can wait for more than 5m if they want.

Robinlovelace (approved):
Looks good, I made a minor change and have an idea to could make 300 s less hardcoded. That could make it easier for people with v. slow connections to temporarily go above 5 minutes for download wait times.

Overall: v. happy with this, great work Andrea!

R/download.R
agila5: Thanks for the reviews and the fix. 

> Looks good, just tested and I get:

I think that the important part is that `getOption()` returns 300 after running that code: 

``` r
getOption("timeout")
#> [1] 60
res = options(timeout = max(300, getOption("timeout")))
getOption("timeout")
#> [1] 300
res$timeout
#> [1] 60
```

<sup>Created on 2021-08-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

> Idea: add an argument timeout so the user can wait for more than 5m if they want. 

IMO no since we have tons of arguments already and if you really need more than 5 mins (not sure if it's that common), then you can set it before running `oe_get`. For example: 

``` r
options(timeout = 600)
getOption("timeout")
#> [1] 600
res = options(timeout = max(300, getOption("timeout")))
getOption("timeout")
#> [1] 600
res$timeout
#> [1] 600
```

<sup>Created on 2021-08-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

Robinlovelace: > IMO no since we have tons of arguments already and if you really need more than 5 mins (not sure if it's that common), then you can set it before running `oe_get`. For example:

Agreed :+1:

R/download.R
```diff
@@ -133,6 +133,9 @@ oe_download = function(
       stop("Aborted by user.")
     }
 
+    old_opts = options(timeout = max(300, getOption("timeout")))
+    on.exit(options(old_opts))
```
hrodmn, line 137: Hello! I think the new timeout behavior is not working exactly as intended. I have a script where I set the `timeout` option to 1000, but my call to `osmextract::oe_download` is timing out at 300 seconds. This code used to work for me on an older version of `osmextract`, but now it won't work:
```r
# update to avoid download timeout
options(timeout = 1000)

# fix certificates
httr::set_config(httr::config(ssl_verifypeer = 0L))

# usa osm.pbf
usa_pbf_url <- "http://download.geofabrik.de/north-america/us-latest.osm.pbf"

# download
usa_pbf <- osmextract::oe_download(usa_pbf_url)
```

agila5: Hi @hrodmn! Yes, you are right, we switched from `download.file` to `httr::GET` a few months ago: 

https://github.com/ropensci/osmextract/blob/f1cccc719a9106ac3b8af623f42c406880cef59f/R/download.R#L138-L147

Would you mind creating a new issue [here](https://github.com/ropensci/osmextract/issues) ?

agila5: I think I fixed the problem in https://github.com/ropensci/osmextract/commit/182506ffb57960b79ec1fa4e23f45743f01df6d9 Please test the github version of the package.

PR #224, opened 2021-08-24
Add new types to oe_get_network

agila5:
New definitions are taken from [osminfra](https://github.com/Robinlovelace/osminfra) package. For example: 

``` r
library(osmextract)
library(sf)
#> Linking to GEOS 3.9.0, GDAL 3.2.1, PROJ 7.2.1

iow <- oe_get_network("isle of wight", "cycle_infrastructure")
#> The input place was matched with: Isle of Wight
#> The chosen file was already detected in the download directory. Skip downloading.
#> Start with the vectortranslate operations on the input file!
#> 0...10...20...30...40...50...60...70...80...90...100 - done.
#> Finished the vectortranslate operations on the input file!
#> Reading layer `lines' from data source 
#>   `C:\Users\Utente\Documents\osm-data\geofabrik_isle-of-wight-latest.gpkg' 
#>   using driver `GPKG'
#> Simple feature collection with 6345 features and 17 fields
#> Geometry type: LINESTRING
#> Dimension:     XY
#> Bounding box:  xmin: -1.584214 ymin: 50.57523 xmax: -1.069921 ymax: 50.76732
#> Geodetic CRS:  WGS 84
plot(iow["highway"], key.length = 1, main = "", key.pos = 1)
```

<sup>Created on 2021-08-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

I'm not sure about all details behind those definitions, but, if you want, I can complete the PR and add all new types. 

cc @Robinlovelace and @joeytalbot

### DIFF
--- a/R/get-network.R
+++ b/R/get-network.R
@@ -85,7 +85,7 @@
 #' plot(its_driving["highway"], lwd = 2, key.pos = 4, key.width = lcm(2.75))
 oe_get_network = function(
   place,
-  mode = c("cycling", "driving", "walking"),
+  mode = c("cycling", "driving", "walking", "cycle_infrastructure"),
   ...
 ) {
   # Load the relevant oe_get options
@@ -94,7 +94,8 @@ oe_get_network = function(
     mode,
     cycling = load_options_cycling(place),
     walking = load_options_walking(place),
-    driving = load_options_driving(place)
+    driving = load_options_driving(place),
+    cycle_infrastructure = load_options_cycle_infrastructure(place)
   )
 
   # Check the other arguments supplied by the user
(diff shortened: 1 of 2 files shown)


Robinlovelace:
This looks good to me, I think it makes sense to have it in there. Note, we'll probably need to change the function, specifically, the inclusion of footway was not right. That will be fixed in this small change: https://github.com/ropensci/osmextract/pull/225

Robinlovelace:
Note the updated definitions here, there are many different ways to define a cycleway!

https://github.com/Robinlovelace/osminfra/blob/main/R/get.R

agila5:
> This looks good to me, I think it makes sense to have it in there. Note, we'll probably need to change the function, specifically, the inclusion of footway was not right. That will be fixed in this small change: #225

👍 

> Note the updated definitions here, there are many different ways to define a cycleway!

I think it should be pretty easy to copy those definitions here and adjust the docs (with links maybe) documenting the new types.

PR #243, opened 2022-02-06
Conditionally import `...names()`

### LINKED ISSUE #241: oe_get - Error in ...names() : could not find function "...names"
Hi! 
I want to use OSM data using R and I found this wonderful package `osmextract`. But when I run a simple command to download all polylines available in OSM for India, I get this error shown below. Any help is appreciated, thanks! (This error comes up even when I try to use the examples shown in the [README ](https://github.com/ropensci/osmextract)of `osmextract`). I saw [this](https://github.com/ropensci/osmextract/issues/234), but sorry could not understand much. 

```{r}
library(sf)
library(osmextract)
oe_get(place = "India", layer = "lines", provider = "geofabrik", match_by = "name") 
# The input place was matched with: India
# Error in ...names() : could not find function "...names"
```

agila5:
Fix #241. 

The problem is that the `...names()` function was introduced in `R` v4.1 while we condition only on `R` >= 3.5. It is quite difficult to reimplement the existing code without `...names()`, so I decided to add a new dependency (`backports`) and conditionally import `...names()` when R < 4.1.

### DIFF
--- a/R/get-key-values.R
+++ b/R/get-key-values.R
@@ -6,32 +6,36 @@
 #' stored first.
 #'
 #' @details OSM data are typically documented using several
-#'   [`tags`](https://wiki.openstreetmap.org/wiki/Tags), i.e. pairs of two
-#'   items, namely a `key` and a `value`. The conversion between `.osm.pbf` and
-#'   `.gpkg` formats is governed by a `CONFIG` file that lists which tags must
-#'   be explicitly added to the `.gpkg` file. All the other keys are
-#'   automatically stored using an `other_tags` field with a syntax compatible
-#'   with the PostgreSQL HSTORE type. See
-#'   [here](https://gdal.org/drivers/vector/osm.html#driver-capabilities) for
-#'   more details.
+#' [`tags`](https://wiki.openstreetmap.org/wiki/Tags), i.e. pairs of two
+#' items, namely a `key` and a `value`. The conversion between `.osm.pbf` and
+#' `.gpkg` formats is governed by a `CONFIG` file that lists which tags must
+#' be explicitly added to the `.gpkg` file. All the other keys are
+#' automatically stored using an `other_tags` field with a syntax compatible
+#' with the PostgreSQL HSTORE type. See
+#' [here](https://gdal.org/drivers/vector/osm.html#driver-capabilities) for
+#' more details.
 #'
-#'   When the argument `values` is `TRUE`, then the function returns a named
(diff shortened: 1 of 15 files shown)


agila5:
Unfortunately, we need to wait for https://github.com/r-lib/backports/issues/65 (or, maybe, I will just remove those tests)

agila5:
Unfortunately that still doesn't work, see the issue in the backports repo...

agila5:
At the end, I reimplemented the same approach using `names(list(...))` and `tryCatch` (i.e. without `backports` or `...names()`).

R/get-key-values.R
```diff
@@ -62,7 +61,7 @@
 #' @examples
 #' # Copy ITS file to tempdir so that the examples do not require internet
 #' # connection. You can skip the next few lines (and start directly with
-#' # oe_get_keys) when running the examples ocally.
+#' # oe_get_keys) when running the examples locally.
```
Robinlovelace, line 65: :+1:

R/get-key-values.R
```diff
@@ -105,13 +104,25 @@
 #' # Remove .pbf and .gpkg files in tempdir
 #' # (since they may interact with other examples)
 #' file.remove(list.files(path = tempdir(), pattern = "(pbf|gpkg)", full.names = TRUE))
-oe_get_keys = function(zone, layer = "lines", values = FALSE, which_keys = NULL) {
+oe_get_keys = function(
+  zone,
```
Robinlovelace, line 109: Really like this style, with each argument on a new line.

R/get-key-values.R
```diff
@@ -191,28 +208,25 @@ oe_get_keys.character = function(zone, layer = "lines", values = FALSE, which_ke
...
-      quiet = TRUE
-    )
-  } else {
-    obj = sf::st_read(
-      dsn = zone,
-      query = paste0("select other_tags from ", layer),
-      quiet = TRUE
-    )
-  }
+  obj = my_st_read(
```
Robinlovelace, line 212: :eyes:

Robinlovelace: Looking at this :+1: 

https://github.com/ropensci/osmextract/blob/remove-.names/R/utils.R#L31-L57

Robinlovelace (approved):

Robinlovelace:
Cannot see the new dep on backports and TBH not 100% sure what is going on. My general thought is that there's no harm in insisting on recent R versions. But if it works and enables more people to use the package, as seems to be the case, :+1: from me, as per approval in review.

PR #271, opened 2023-02-06
Update geofabrik_zones

### LINKED ISSUE #270: Mismatch between GEOMETRY in `osmextract::geofabrik_zones` and actual clipping boundary after update
Hi

I noticed a small mistake in the osm layer for Belgium from [geofabrik](https://download.geofabrik.de/europe/belgium.html). The osm layer was not completely encompassing Belgium. 

This meant that when I wanted to extract information from a region containing the mismatched area (using `oe_get()`), not the region of Belgium was downloaded but the entire European map. This map is very large and way too large for my intentions. Belgium is only a small country so it is a bit silly to have to download the entire map of Europe. 

The people from geofabrik promptly fixed this, but `oe_get()` (via `oe_match()`) still wants to use the whole of Europe ...
I discovered that this is because the GEOMETRY in `osmextract::geofabrik_zones` is not up to date with the actual clipping boundary of the osm layer on the geofabrik website.  The clipping boundaries are available as *country_name*.kml files on [https://download.geofabrik.de/europe/](https://download.geofabrik.de/europe/).

Therefore my question:
Can you update `osmextract::geofabrik_zones` GEOMETRY such that it is up to date with the actual layers behind the download links?
 
Or even better:
Would it be possible to automate this process? If there is an update from geofabrik (or any other provider), the GEOMETRY is updated automatically as well. 

Kind regards,
Ward

agila5:
Fix #270. See `geofabrik_zones.R` to check the differences between the old and new approaches.


agila5:
Please note that the different number of rows (described in the .Rd file) is due to some extracts created just recently, e.g. https://download.geofabrik.de/central-america/costa-rica.html.

PR #300, opened 2024-11-01
Fix sql

### LINKED ISSUE #298: Unexpected error when using GDAL 3.10.0rc1 [BUG]
With released GDAL 3.9.3, no error occurs in `oe_get_network`: 
[osmextract.Rcheck.GDAL393.zip](https://github.com/user-attachments/files/17586527/osmextract.Rcheck.GDAL393.zip)
However, when running the same CMD check with the same platform and the same `sf`, GDAL 3.10.0rc1 shows a failure with no `its_walking` features found: 
[osmextract.Rcheck.GDAL3100.zip](https://github.com/user-attachments/files/17586542/osmextract.Rcheck.GDAL3100.zip)
Please run your own checks with a GDAL 3.10.0 release candidate (currently rc2, probably the last) and see whether this is a problem. @agila5 @edzer

agila5:
Fix SQL syntax following the suggestions in #298 for GDAL 3.10. 

Fixes #298.

### DIFF
--- a/R/get-network.R
+++ b/R/get-network.R
@@ -144,6 +144,12 @@ oe_get_network = function(
 # accepted but have support among many mappers. More precise options are defined
 # by other tags. See also: https://wiki.openstreetmap.org/wiki/Bicycle#Bicycle_Restrictions
 
+# WARNING: Starting from GDAL 3.10, an expression like "foo NOT IN ('bar')"
+# evaluates as false, while previously it would evaluate as true. Therefore, in
+# the following code, when building the -where clause, we need to be explicit
+# and include possible NULL values in the expression. See also the discussion in
+# #298 and the fix implemented by @rouault.
+
 # A cycling mode of transport includes the following scenarios:
 # - highway IS NOT NULL (since usually that means that's not a road);
 # - highway NOT IN ('abandoned', 'bus_guideway', 'byway', 'construction', 'corridor',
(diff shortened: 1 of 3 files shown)


R/get-network.R
```diff
@@ -144,6 +144,12 @@ oe_get_network = function(
 # accepted but have support among many mappers. More precise options are defined
 # by other tags. See also: https://wiki.openstreetmap.org/wiki/Bicycle#Bicycle_Restrictions
 
+# WARNING: Starting from GDAL 3.10, an expression like "foo NOT IN ('bar')"
+# evaluates as false, while previously it would evaluate as true. Therefore, in
```
Robinlovelace, line 148: Is there a link to the GDAL documentation that says something similar?

Robinlovelace (commented):
Looks like progress, wonder if it passes..

R/get-network.R
rouault: yes see https://github.com/OSGeo/gdal/blob/master/MIGRATION_GUIDE.TXT#L4

Robinlovelace: Great, let's add that link to the docs.

rouault: (suggestion: use a permalink or link to the tag: https://github.com/OSGeo/gdal/blob/v3.10.0RC3/MIGRATION_GUIDE.TXT)

agila5: Added permalink, thank you very much!

PR #310, opened 2025-07-11
Change the default download directory from tempdir to R_user_dir("osmextract", "data")

### LINKED ISSUE #302: [FEATURE] Replace tempdir() with R_user_dir()
The R 4.0.0 NEWS file says: 

>  New function R_user_dir() in package tools suggests paths appropriate for storing R-related user-specific data, configuration and cache files.

For example, I currently see: 

``` r
`tools::R_user_dir("osmextract", "data")`
#> [1] "C:\\Users\\user\\AppData\\Roaming/R/data/R/osmextract"
```

I think that the ideas behind this function are somewhat related to the approach we currently use to detect where `osmextract` functions will save `.osm` and `.gpkg` files, i.e. 

https://github.com/ropensci/osmextract/blob/0eeacf8c725d66a93fec6e8ef9112178e6071db2/R/utils.R#L72-L78

Currently, we set the download directory as `tempdir()` whenever the `OSMEXT_DOWNLOAD_DIRECTORY` envvar is not set. 

Does it make sense switching to `tools::R_user_dir("osmextract", "data")`?

agila5:
Fixes #302.


agila5:
Hi Robin! Can you please check here, especially the description included in the README file? The old comments regarding tempdir do not make sense any more, and I'm not sure whether the new wording is appropriate. Thanks 🙇‍♂️.

Robinlovelace:
I've had a quick read and the new wording is perfect. Will this prevent the situation in which identical pbf files are downloaded multiple times for users who didn't set the download directory? That's most users so thus will save internet resources!

agila5:
> I've had a quick read and the new wording is perfect. Will this prevent the situation in which identical pbf files are downloaded multiple times for users who didn't set the download directory? That's most users so thus will save internet resources!

Hopefully yes. I'm not sure about unintended consequences, but I guess we'll see 😅

Issue #320, opened 2026-06-17
[FEATURE] Process the `oneway` tag


juanfonsecaLS1:
**Is your feature request related to a problem? Please describe.**

Currently, `oe_get_network` extracts raw OSM tag values for `oneway` as they are. This results in a mix of values like: NA, "yes", "no", "alternating", "-1", and "reversible", which can be problematic for using the data in `dodgr` for example.

Furthermore, `junction=roundabout` features are implicitly one-way, but since they sometimes lack an explicit `oneway` tag, they are not identified as such. As mentioned in #319 

See [Wiki](https://wiki.openstreetmap.org/wiki/Tag:junction%3Droundabout)

**Describe the solution you'd like**
I would like `oe_get_network` to standardise the output of the `oneway` column to a consistent format (such as `yes/no` or boolean `TRUE/FALSE`). Also reversing the geometries  with `-1`.
Also, the function there could be a check so if `junction` is set to `roundabout` or `circular` and `oneway` is not explicitly set, it should automatically resolve to `yes` (or `TRUE`).

**Additional context**

An example of the different values for `oneway`:

``` r
library(sf)
#> Linking to GEOS 3.14.1, GDAL 3.12.1, PROJ 9.7.1; sf_use_s2() is TRUE
library(osmextract)
#> Data (c) OpenStreetMap contributors, ODbL 1.0. https://www.openstreetmap.org/copyright.
#> Check the package website, https://docs.ropensci.org/osmextract/, for more details.
library(tidyverse)

data <- oe_get_network("West Yorkshire",mode = "driving")
#> The input place was matched with: West Yorkshire
#> The chosen file was already detected in the download directory. Skip downloading.
#> Starting with the vectortranslate operations on the input file!
#> 0...10...20...30...40...50...60...70...80...90...100 - done.
#> Finished the vectortranslate operations on the input file!
#> Reading layer `lines' from data source 
#>   `C:\temp\osmextract\geofabrik_west-yorkshire-latest.gpkg' using driver `GPKG'
#> Simple feature collection with 134838 features and 14 fields
#> Geometry type: LINESTRING
#> Dimension:     XY
#> Bounding box:  xmin: -2.177229 ymin: 53.52073 xmax: -1.162529 ymax: 53.96817
#> Geodetic CRS:  WGS 84

data |> pull(oneway) |> unique()
#> [1] NA            "yes"         "no"          "alternating" "-1"         
#> [6] "reversible"
```

[...]

PR #321, opened 2026-07-02
added functions for obtaining graphs from osm 

juanfonsecaLS1:
- **new network graph functions**
- **included brief description of new functionalities in NEWS**
- **fixed validation**
- **improved the wording in NEWS**
- **added tests for the new functions**
- **added test for dodgr**
- **updated documentation**

## Description

This adds some wrappers to obtain fully routable networks with minimal pre-processing based on the geometries and the attributes, further network simplification is left to the user.

This is equivalent to what you would get with the `graph_from_*` in `osmnx`.

The naming of the functions might need to be reviewed, for example `oe_get_sfnetwork` or `oe_get_network_sfnet`...?

## Related Issue
This is related to issue #320. And adds new functionalities mentioned in the same thread

## Example

Using the ITS Leeds data we can get:

- A tidy `sf` object with yes/no values for oneway and coherent geometries and tidy highway values (no _link)
oe_get_tidynetwork(
   place = "ITS Leeds",
   mode = "driving",
   highway_filter = highway_filter
)

- A `sfnetwork` with a undirected/directed graph
oe_get_sfnetwork(
   place = "ITS Leeds",
   mode = "walking",
   directed = FALSE
)

- Or a `dodgr_streetnet`: 
oe_get_dodgrnetwork(
    place = "ITS Leeds",
    mode = "cycling",
    wt_profile = "bicycle",
    left_side = TRUE,
    highway_filter = highway_filter
)

Some other helper functions have been added but not included in the exports.

Tests have been added for all functions and arguments.


R/get-network-graph.R
```diff
@@ -0,0 +1,401 @@
...
+#' plot(walk_sfnet)
+#'
+
+oe_get_sfnetwork <- function(
+  ...,
+  directed = FALSE,
+  simplify_highway = TRUE,
+  highway_filter = NULL
+) {
+  rlang::check_installed(
```
agila5: I wouldn't add rlang as a dependency just to check whether these two packages are available. You can add something like

```
if (!requireNamespace("tidygraph", quietly = TRUE)) stop("sfnetworks is not available. Please install it first")
if (!requireNamespace("sfnetworks", quietly = TRUE)) stop("sfnetworks is not available. Please install it first")
```

R/get-network-graph.R
```diff
@@ -0,0 +1,401 @@
...
+#'
+#' This function is a wrapper around `oe_get_network()` that returns a `sfnetwork` object.
+#' It performs simplification of the highway values and filters by highway types if specified. Minimal
+#' network preprocessing tasks i.e. subdivision and smoothing are performed using `sfnetworks` to
+#' create a tidy `sfnetwork` object. All unique merged edge attributes are concatenated.
+#' It also allows for the creation of directed or undirected networks.
+#'
+#' @param ... parameters passed to `oe_get_network()`
+#' @param simplify_highway logical, whether to simplify the highway values by removing the "_link" suffix and filtering by `highway_filter`
+#' @param highway_filter character vector of highway types to keep, if `simplify_highway` is TRUE
```
agila5: I see that the default value is NULL. Can you specify (here on in other parts of the docs) what happens when we keep the NULL default?

R/get-network-graph.R
```diff
@@ -0,0 +1,401 @@
...
+#' print(tidynet_sf)
+#'
+#' tidynet_sf_filtered <- oe_get_tidynetwork(
+#'   place = "ITS Leeds",
+#'   mode = "driving",
+#'   highway_filter = highway_filter
+#' )
+#' print(tidynet_sf_filtered)
+#'
+oe_get_tidynetwork <- function(
```
agila5: I don't understand the objective of this function. It should be oe_get_network + highway filtering and preprocessing? In that case I would add those parameters to the original oe_get_network. 

@Robinlovelace what do you think?

R/get-network-graph.R
```diff
@@ -0,0 +1,401 @@
...
+#' @param highway_filter character vector of highway types to keep, if `simplify_highway` is TRUE
+#' @param directed logical, whether to return a directed sfnetwork object (default is FALSE)
+#'
+#' @returns a `sfnetwork` object
+#'
+#' @export
+#'
+#' @seealso [oe_get()], [oe_get_network()], [oe_get_tidynetwork()], [oe_get_dodgrnetwork()]
+#' @examples
+#' highway_filter = c(
```
agila5: Can you write in the docs which are the valid values for the highway filter?

R/get-network-graph.R
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+  net <- oe_get_tidynetwork(
+    ...,
+    simplify_highway = simplify_highway,
+    highway_filter = highway_filter
+  )
+
+  # Basic simplification using sfnetworks with the undirected graph
+  message("Starting basic network simplification...")
```
agila5: Please add a quiet argument to the function (default value is usually FALSE) and use something like

```
oe_message(
    "Starting basic network simplification..., 
    quiet = quiet,
    .subclass = "oe_get_sfnetwork_simplification"
)
```

R/get-network-graph.R
```diff
@@ -0,0 +1,401 @@
...
+#'    wt_profile = "bicycle",
+#'    left_side = TRUE,
+#'    highway_filter = highway_filter
+#'  )
+#'
+oe_get_dodgrnetwork <- function(
+  ...,
+  highway_filter = NULL
+) {
+  rlang::check_installed(
```
agila5: I guess that there is a typo (i.e. sfnetworks should be dodgr) and, additionally, I would use the coding pattern mentioned before (i.e. !requireNamespace ...)

R/get-network-graph.R
```diff
@@ -0,0 +1,401 @@
...
+
+  # Reversing the geometries with -1
+  sf::st_geometry(net_raw[
+    net_raw$oneway == "-1",
+  ]) <- sf::st_reverse(sf::st_geometry(net_raw[net_raw$oneway == "-1", ]))
+
+  net_raw$oneway[net_raw$oneway == "-1"] <- "yes"
+
+  if ("junction" %in% names(net_raw) && implied_oneway) {
+    message("The implied oneway restriction is applied.")
```
agila5: Please add a quiet argument and use oe_message following the same pattern described above

tests/testthat/test-get-network-graph.R
```diff
@@ -0,0 +1,153 @@
+test_that("tidy_highway strips _link suffix and filters values", {
```
agila5: These tests should be run conditionally on the presence of sfnetworks / dodgr / tidygraph. 

Please add something like `skip_if_not_installed("sfnetworks")` at the beginning of these tests

agila5 (changes requested):

R/get-network-graph.R
juanfonsecaLS1: Yes, agree. I will change it. The only reason I kept `rlang` was to make it easy for users to install packages on the go.

R/get-network-graph.R
juanfonsecaLS1: I will add thesse details in the description of the parameter.

R/get-network-graph.R
juanfonsecaLS1: I will check `sfnetworks` first. So the user only runs one installation as `tidygraph` is a dependency of it

R/get-network-graph.R
juanfonsecaLS1: fixed, also I will pass the quiet argument to the `oe_get_network` call

R/get-network-graph.R
juanfonsecaLS1: Correct. I fixed and implemented the check as in the previous instance

R/get-network-graph.R
juanfonsecaLS1: fixed

tests/testthat/test-get-network-graph.R
juanfonsecaLS1: fixed, added `sfnetworks` and `dodgr` where needed

R/get-network-graph.R
juanfonsecaLS1: Fixed, the parameter is better documented now

juanfonsecaLS1:
Hi @agila5, thanks for your comments, I have addressed all of them but the `oe_get_tidynetwork` one.

agila5:
Awesome, thanks. I'll review everything again tomorrow.

R/get-network-graph.R
```diff
@@ -13,9 +14,9 @@
 #' "path", "pedestrian", "primary", "residential", "rest_area", "service", "services",
 #' "steps", "tertiary", "track", "trunk" and "unclassified".
 #' @param directed logical, whether to return a directed sfnetwork object (default is FALSE)
-#' @inheritParams oe_get quiet
+#' @inheritParams oe_get
 #'
-#' @returns a `sfnetwork` object
+#' @returns A `sfnetwork` object
```
Robinlovelace:
```suggestion
#' @returns An `sfnetwork` object
```

https://www.quora.com/When-a-word-starts-with-an-%E2%80%9Cs%E2%80%9D-do-you-use-%E2%80%9Ca%E2%80%9D-or-%E2%80%9Can%E2%80%9D-before-the-word?__cf_chl_f_tk=0VjKsay7fkM8oz7Sh.83q9u_CF0axLzmUNHWHMPaRk0-1783076089-1.0.1.1-_rvRrUKrHkhRH5PBTAQm8uSBBdQqPbWFC_pTF5LqSPU

R/get-network-graph.R
Robinlovelace: Happy for it to be moved into `oe_get_network` if useful for most users.

Robinlovelace (approved):

R/get-network-graph.R
juanfonsecaLS1: should I add a logical parameter like `tidy_output`, set as `FALSE` by default? AND if `TRUE`, it triggers the code for `oneway` + `highway`

agila5: Yes, please! Then I will finish the PR and merge!

agila5: Maybe clean_output instead of tidy_output?

juanfonsecaLS1: Yes, agree. Using tidy is misleading. I will push the changes soon.

juanfonsecaLS1:
@agila5 this is ready for you to review, as discussed I added new parameters in the `oe_get_network`. Some changes were introduced by air (the formatter), e.g. line breaks in some of the tests and in the definition of the sql query for the different modes.

agila5:
Awesome, thanks! I'm not sure I can check it today but, worst case scenario, I'll review it on Sunday or Monday. Thanks again for your work!

R/get-network-graph.R
```diff
@@ -0,0 +1,355 @@
...
+
+oe_get_sfnetwork = function(
+  place,
+  mode = c("cycling", "driving", "walking"),
+  ...,
+  directed = FALSE,
+  highway_filter = NULL,
+  quiet = FALSE
+) {
+  if (!requireNamespace("sfnetworks", quietly = TRUE)) {
```
agila5: Two comments: 

* tidygraph is a strong dependency of sfnetworks, so it doesn't make sense to check its presence after loading sfnetworks since, in that case, you are not even able to load sfnetworks. For example: 

``` r
remove.packages("tidygraph")
#> Removing package from 'C:/Users/andre/AppData/Local/R/win-library/4.4'
#> (as 'lib' is unspecified)
#> Error in `find.package()`:
#> ! there is no package called 'tidygraph'
requireNamespace("sfnetworks")
#> Loading required namespace: sfnetworks
#> Failed with error:  ‘there is no package called ‘tidygraph’’
```

<sup>Created on 2026-07-05 with [reprex v2.1.1.9000](https://reprex.tidyverse.org)</sup>

* I noticed that tidygraph and sfnetworks are not even used in this function so it's not immediately clear why we are testing their presence. I would move such namespace tests to the appropriate functions (i.e. net_2_sfnet_undirected and so on).

R/get-network-graph.R
```diff
@@ -0,0 +1,355 @@
...
+  }
+  if (!requireNamespace("tidygraph", quietly = TRUE)) {
+    stop("tidygraph is not available. Please install it first")
+  }
+
+  if (!is.null(highway_filter)) {
+    check_highway_filter(highway_filter)
+  }
+
+  if (!is.logical(directed) || length(directed) != 1) {
```
agila5: IMO, these type of tests can be written more cleanly as 

```
stopifnot(is.logical(directed) && length(directed) == 1L)
```

Feel free to ignore this comment if you prefer the previous syntax.

R/get-network-graph.R
```diff
@@ -0,0 +1,355 @@
...
+
+  # Basic simplification using sfnetworks with the undirected graph
+
+  oe_message(
+    "Starting basic network pre-processing...",
+    quiet = quiet,
+    .subclass = "oe_get_sfnetwork_simplification"
+  )
+
+  net = net_2_sfnet_undirected(net)
```
agila5: I don't understand why the sfnetworks steps are always run on an undirected network and, after that, we convert it into a directed one. I'm not sure that the two procedures (i.e. morphers on undirected --> directed vs morphers on directed) are equivalent. Is there any particular reason why you decided to write the code this way? Am I missing something?

R/get-network-graph.R
```diff
@@ -0,0 +1,355 @@
...
+  # Returning the weighted_streetnetwork
+  do.call(dodgr::weight_streetnet, dodgr_args)
+}
+
+
+## Utils
+
+# This function simplifies the highway values by removing the "_link" suffix and filtering by `highway_filter` if specified.
+
+tidy_highway = function(net, highway_filter) {
```
agila5: Please add here a line which tests that the `highway` column is present in the data. If not, I would raise a warning and return net. Something like: 

``` r
if (! "highway" %in% colnames(net)) {
  warning("The highway key is not present in the data. Skipping cleaning ops.")
  return(net)
}
```

R/get-network.R
```diff
@@ -9,6 +9,13 @@
 #' @param mode A character string of length one denoting the desired mode of
 #'   transport. Can be abbreviated. Currently `cycling` (the default), `driving`
 #'   and `walking` are supported.
+#' @param clean_output logical, whether to standardise `oneway`
```
agila5: If I read the code correctly, `tidy_oneway()` also reverses the wrong-way roads, right? If so, please add a sentence here detailing such behaviour.

R/get-network-graph.R
```diff
@@ -0,0 +1,355 @@
...
+#'
+#' @details For more information on the implied oneway restriction, see [wiki.openstreetmap.org](https://wiki.openstreetmap.org/wiki/Key:oneway#Implied_oneway_restriction).
+#'
+#'
+#' @examples
+#' \dontrun{
+#' sf_net <- osmextract::oe_get_network(place = "ITS Leeds", mode = "driving")
+#' sf_net_tidy <- tidy_oneway(sf_net)
+#' }
+tidy_oneway = function(
```
agila5: What do you think about using `clean_highway` and `clean_oneway` instead of `tidy_highway` and `tidy_oneway`?

R/get-network-graph.R
```diff
@@ -0,0 +1,355 @@
...
+
+  if ("junction" %in% names(net_raw)) {
+    oe_message(
+      "The implied oneway restriction was applied based on the junction column values.",
+      quiet = quiet,
+      .subclass = "tidy_oneway_implied"
+    )
+
+    net_raw$oneway[
+      net_raw$junction %in% c("roundabout", "motorway") & net_raw$oneway == "no"
```
agila5: Are you sure that we should set `oneway = 'yes'` even when the original data explicitly reported `oneway = 'no'`? Is there a particular reason that I'm missing?

R/get-network.R
```diff
@@ -130,12 +139,50 @@ oe_get_network = function(
...
+    # Simplifies the highway values by removing the "_link" suffix
+    result = tidy_highway(result, highway_filter = highway_filter)
+
+    # Check if the oneway column is present, and if not,
+    # adds it with the default value "no"
+    if ("oneway" %in% names(result)) {
+      result = tidy_oneway(result, quiet = quiet)
+    } else {
+      oe_message(
+        "The 'oneway' column is missing. 'no' will be used as the default value.",
```
agila5: Is there any particular reason why we forcefully add a oneway column with no as a default value? Wouldn't be better not to add any artificial column to the data?

agila5:
Hi @juanfonsecaLS1 and thanks again for your work. I added a few more questions on the PR. I apologise that I'm being very strict, but I'd love these functions to be implemented as good as possible. Feel free to just answer my questions and I can complete the coding parts if you want.

juanfonsecaLS1:
Hi @agila5, no problem at all! I understand. I am also interested in not missing anything for these functions.
Although, I cannot see any new comments on the PR.

Robinlovelace:
> I added a few more questions on the PR

Hey @agila5 what are the questions? We cannot see them...

R/get-network-graph.R
```diff
@@ -0,0 +1,355 @@
...
+
+  net_raw$from = NULL
+  net_raw$to = NULL
+  net_raw$z_order = NULL
+
+  # Reversing the geometries of bidirectional links
+  net_rev = sf::st_reverse(net_raw[net_raw$oneway == "no", ])
+
+  # Binding the duplicated geometries
+  rbind(net_rev, net_raw) |>
```
Robinlovelace: Heads-up that this can be slow for large network.

Robinlovelace: One idea: output a message saying install dplyr for faster. Could do if statement with package namespace and use dplyr if available?

R/get-network-graph.R
```diff
@@ -0,0 +1,355 @@
...
+    x = net_sf,
+    directed = FALSE
+  )
+
+  # Creating junctions where road segments overlap
+  # This converts implicit intersections into explicit nodes
+  sf_net_subdiv = tidygraph::convert(sfnet, sfnetworks::to_spatial_subdivision)
+
+  # Simplifying the interstitial nodes segments keeping
+  # the oneway attribute and concatenating the other fields
```
Robinlovelace:
```suggestion
  # the oneway attribute and concatenating the other fields
  # TODO: think about ways to speed-up, e.g. if graph above certain size add message saying maybe slow
```

Robinlovelace (approved):

agila5 (changes requested):

agila5:
I forgot to submit the review, sorry. You should see the Q now.

R/get-network-graph.R
Robinlovelace: Good comments Andrea, thanks 👍

juanfonsecaLS1:
> I forgot to submit the review, sorry. You should see the Q now.

No worries. Thanks @agila5 for this. I will take a look now.

R/get-network.R
juanfonsecaLS1: Yes, correct. I will add the clarification.

R/get-network-graph.R
juanfonsecaLS1: Corrected, I included a check for `sfnetworks` in the two related functions, and skipped the redundant check for `tidygraph`.

R/get-network-graph.R
juanfonsecaLS1: If we add this, should we add `dplyr` in suggests?

R/get-network-graph.R
juanfonsecaLS1: I agree. I will make them consistent, i.e., `clean_*`

R/get-network-graph.R
juanfonsecaLS1: I did actually have an implementation of the morphers on the directed graph, but I found that the results were not what I was expecting. 

The main reason is that the second morpher (to_spatial_smooth) simplifies nodes with `degree = 2`; in the directed graph, those nodes are actually `degree = 4` and are not simplified. In the current implementation, I am ensuring that a simplified geometry have the same direction, all other attributes are collapsed.

This is a minimal example:

``` r
library(sf)
#> Linking to GEOS 3.14.1, GDAL 3.12.1, PROJ 9.7.1; sf_use_s2() is TRUE
library(tidygraph)
#> 
#> Attaching package: 'tidygraph'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(sfnetworks)


# Building a dummy example with 2 `dummy/interstitial` nodes and 3 edges.

p1 = st_point(c(0, 1))
p2 = st_point(c(1, 1))
p3 = st_point(c(2, 1))
p4 = st_point(c(3, 1))
p5 = st_point(c(4, 1))


l1 = st_sfc(st_linestring(c(p1, p2, p3)))
l2 = st_sfc(st_linestring(c(p3, p4)))
l3 = st_sfc(st_linestring(c(p4, p5)))


lines = st_as_sf(x = c(l1, l2, l3))
lines$oneway = c("no", "no", "no")

# Creating an undirected version of the network
sfnet <- as_sfnetwork(lines, directed = FALSE)

# Creating the explicitly directed version of the network
lines_directed <- rbind(lines, st_reverse(lines))
sfnet_directed <- as_sfnetwork(lines_directed, directed = TRUE)

# Simplifying the networks by removing the `dummy/interstitial` nodes
sfnet_smooth <- tidygraph::convert(sfnet, to_spatial_smooth)
sfnet_directed_smooth <- tidygraph::convert(sfnet_directed, to_spatial_smooth)

# plot(sfnet)
# plot(sfnet_smooth)

# The results
igraph::E(sfnet)
#> + 3/3 edges from 41a6603:
#> [1] 1--2 2--3 3--4
igraph::E(sfnet_smooth)
#> + 1/1 edge from 41ad1f1:
#> [1] 1--2
igraph::E(sfnet_directed_smooth)
#> + 4/4 edges from 41b0f93:
#> [1] 1->2 2->1 1->1 2->2
# plot(sfnet_directed_smooth)
```

<sup>Created on 2026-07-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/get-network-graph.R
juanfonsecaLS1: corrected now

R/get-network-graph.R
juanfonsecaLS1: You are right, that can produce errors. According to the Wiki, both `junction = roundabout` and `highway=motorway` imply `"oneway = "yes"` by definition, but we do not know if there are any cases where motorways are represented as a single `LINESTRING`.

As per the Wiki, I will assume those links are oneway = "yes", only if `oneway = NA`. so I will do that first, and then transform the `NA`s into `"no"`.

I just noticed that `motorway` is not a valid value for the`junction` tag. That needs fixing too.

R/get-network-graph.R
juanfonsecaLS1: Oh, it seems better. I will change them. Thanks!

R/get-network-graph.R
juanfonsecaLS1: I will add some code to test the presence of highway.

R/get-network.R
juanfonsecaLS1: Yes, true. My initial thought was to have it available for the graph building part.
But the implementation of the functions for graphs can handle this as: if `oneway` is missing it will force `directed = FALSE`
I will add a line to check if the `oneway` column is present only if `directed = TRUE`

agila5: Thanks! IMO that's a better approach.

R/get-network-graph.R
juanfonsecaLS1: done!

R/get-network-graph.R
juanfonsecaLS1: In the message displayed when the processing starts I have added some text indicating that the process might take a few minutes

juanfonsecaLS1:
@agila5 I think I have addressed all the comments, I have also updated the tests and added a couple for validating the warnings when oneway is missing. I included `dplyr` in Suggests to address Robin's comment. Thanks again for taking the time to review this.

agila5:
Thanks. Checking everything in a few hours!

R/get-network-graph.R
agila5: Ok, I understand. Please add something along these lines to the docs: 

```
#' @details
#' Please notice that the preprocessing tasks are always performed ignoring the
#' directionality of the roads. Then, if `directed = TRUE`, the function will
#' prepare the directed graph by duplicating the bidirectional edges and
#' reversing their geometries. If the `oneway` column is missing, a warning will
#' be issued and an undirected sfnetwork will be returned.
```

R/get-network-graph.R
```diff
@@ -0,0 +1,378 @@
...
+
+  oe_message(
+    "Starting basic network pre-processing. This can take a few minutes for large networks...",
+    quiet = quiet,
+    .subclass = "oe_get_sfnetwork_simplification"
+  )
+
+  net = net_2_sfnet_undirected(net)
+
+  if (directed && !"oneway" %in% names(net)) {
```
agila5: This part doesn't work since `net` is a sfnetwork object and doesn't have names (at least in the regular sense). For example: 

``` r
library(sfnetworks)
as_sfnetwork(roxel) |> names()
#> NULL
```

<sup>Created on 2026-07-07 with [reprex v2.1.1.9000](https://reprex.tidyverse.org)</sup>

so that if clauses is never equal to TRUE

R/get-network-graph.R
```diff
@@ -0,0 +1,378 @@
...
+    directed = FALSE
+  )
+
+  # Creating junctions where road segments overlap
+  # This converts implicit intersections into explicit nodes
+  sf_net_subdiv = tidygraph::convert(sfnet, sfnetworks::to_spatial_subdivision)
+
+  # Simplifying the interstitial nodes segments keeping
+  # the oneway attribute and concatenating the other fields (if oneway is present)
+  if ("oneway" %in% names(sf_net_subdiv)) {
```
agila5: I'm sorry, but I have some doubts regarding this part of the code, especially since it's the default behaviour of the function. This part of the code is extremely intrusive since it strongly modifies the underlying object created by sfnetwork by collapsing the values using ",". For example, it might create _impossible_ highway types such as _primary,secondary_ and so on. 

IMO, we should remove this part of the code and, instead, document this possibility in a vignette or an example. What do you think? Is this workflow something that you usually use in your analysis?

agila5 (changes requested):

R/get-network-graph.R
juanfonsecaLS1: I will add this, including the link to the pre-processing vignette of sfnetworks.

R/get-network-graph.R
juanfonsecaLS1: True. This is too aggressive and can be problematic, especially when a very strict filter is applied, producing continuous *corridors*/`LINESTRINGS` with many different attributes.

The main goal here is to simplify the network graph by removing redundant nodes, which has a direct impact on  entrality calculations, for example.

I would not remove the code completely. I think a better alternative is adding a `require_equal` argument to the function call, to be passed to `to_spatial_smooth`, allowing the user to specify what they want to preserve/collapse based on their use case. We can keep `TRUE` as the default. 

What do you think?

R/get-network-graph.R
juanfonsecaLS1: Fixed

R/get-network-graph.R
agila5: I fear that, when setting `require_equal = TRUE` then, at the end, no redundant node is removed since two edges will never be exactly equal for all possible attributes. Anyway, I guess we can start with your approach (i.e. require_equal = TRUE) and then adjust the function based on additional feedback. Alternatively, we can create a `require_equal` argument (with default value TRUE) which is ultimately passed to to_spatial_smooth.

juanfonsecaLS1:
I just pushed the latest commits with all changes

Robinlovelace:
Any more comments @agila5 ? Many thanks for detailed review, definitely beneficial for code quality! This should be a major release I think.

agila5:
Hi @juanfonsecaLS1 @Robinlovelace, I've checked the PR and I have only a final issue. 

Unfortunately, I noticed that running `oe_get_sfnetworks` on the default ITS data raises a warning regarding attributes that should be constant when subdividing the geometries. For example: 

``` r
library(osmextract)
toy <- oe_get_sfnetwork(
  place = "ITS Leeds",
  mode = "walking",
  download_directory = tempdir(),
  quiet = TRUE
)
#> Warning: to_spatial_subdivision assumes attributes are constant over geometries
```

<sup>Created on 2026-07-08 with [reprex v2.1.1.9000](https://reprex.tidyverse.org)</sup>

This is quite bad since it implies we cannot use the toy ITS data in the tests of the package (otherwise it raises unhandled warnings, which is what's happening right now) and there are minor frictions in the tests. For the example, this is the current output of `devtools::test()`: 

immagine

See also the "Show testthat output" in the GHA. Ideas? I would strongly avoid `suppressWarnings()`.

Robinlovelace:
Maybe we should update the toy example dataset?

agila5:
That's not going to work since `sfnetworks::to_spatial_subdivision` tests the `agr` attribute which are set by `sf` functions and, AFAIK, are not saved when storing data on disk. I'm reading more details right now...

juanfonsecaLS1:
Hi @agila5, this is actually an *expected* warning when you run `to_spatial_subdivision` on most networks. 

It is alerting that a single LINESTRING becomes two or more LINESTRINGS/edges with the same attributes, which is what we want subdivision to do. If no possible subdivisions are found, no warning is raised. 

For example:

``` r
library(sf)
#> Linking to GEOS 3.14.1, GDAL 3.12.1, PROJ 9.7.1; sf_use_s2() is TRUE
library(tidygraph)
#> 
#> Attaching package: 'tidygraph'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(sfnetworks)


# Building a dummy example with 2 `dummy/interstitial` nodes and 3 edges.

p1 = st_point(c(1, 0))
p2 = st_point(c(1, 1))
p3 = st_point(c(1, 2))
p4 = st_point(c(0, 1))
p5 = st_point(c(1, 1))
p6 = st_point(c(2, 1))


l1 = st_sfc(st_linestring(c(p1, p2, p3)))
l2 = st_sfc(st_linestring(c(p4, p5, p6)))

l3 = st_sfc(st_linestring(c(p1, p3)))
l4 = st_sfc(st_linestring(c(p4, p6)))

lines = st_as_sf(x = c(l1, l2))
lines$oneway = c("no", "no")
lines$attr = c("A", "B")

lines2 = st_as_sf(x = c(l3, l4))
lines$oneway = c("no", "no")
lines$attr = c("A", "B")

# Creating the sfnetwork
sfnet <- as_sfnetwork(lines, directed = FALSE)
sfnet_subdivided <- convert(sfnet,to_spatial_subdivision)
#> Warning: to_spatial_subdivision assumes attributes are constant over geometries

# This does not generate the warning as no subdivision happens
sfnet_2 <- as_sfnetwork(lines2, directed = FALSE)
sfnet_2_subdivided <- convert(sfnet_2,to_spatial_subdivision)

# plot(sfnet)
# plot(sfnet_subdivided)

as_tibble(sfnet,"edges")
#> Simple feature collection with 2 features and 4 fields
#> Geometry type: LINESTRING
#> Dimension:     XY
#> Bounding box:  xmin: 0 ymin: 0 xmax: 2 ymax: 2
#> CRS:           NA
#> # A tibble: 2 × 5
#>    from    to               x oneway attr 
#>   <int> <int>    <LINESTRING> <chr>  <chr>
#> 1     1     2 (1 0, 1 1, 1 2) no     A    
#> 2     3     4 (0 1, 1 1, 2 1) no     B
as_tibble(sfnet_subdivided,"edges")
#> Simple feature collection with 4 features and 5 fields
#> Geometry type: LINESTRING
#> Dimension:     XY
#> Bounding box:  xmin: 0 ymin: 0 xmax: 2 ymax: 2
#> CRS:           NA
#> # A tibble: 4 × 6
#>    from    to            x oneway attr  .tidygraph_edge_index
#>   <int> <int> <LINESTRING> <chr>  <chr>                 <int>
#> 1     1     2   (1 0, 1 1) no     A                         1
#> 2     2     3   (1 1, 1 2) no     A                         1
#> 3     2     4   (0 1, 1 1) no     B                         2
#> 4     2     5   (1 1, 2 1) no     B                         2

# It was not subdivided/no warning
as_tibble(sfnet_2,"edges")
#> Simple feature collection with 2 features and 2 fields
#> Geometry type: LINESTRING
#> Dimension:     XY
#> Bounding box:  xmin: 0 ymin: 0 xmax: 2 ymax: 2
#> CRS:           NA
#> # A tibble: 2 × 3
#>    from    to            x
#>   <int> <int> <LINESTRING>
#> 1     1     2   (1 0, 1 2)
#> 2     3     4   (0 1, 2 1)
as_tibble(sfnet_2_subdivided,"edges")
#> Simple feature collection with 2 features and 3 fields
#> Geometry type: LINESTRING
#> Dimension:     XY
#> Bounding box:  xmin: 0 ymin: 0 xmax: 2 ymax: 2
#> CRS:           NA
#> # A tibble: 2 × 4
#>    from    to            x .tidygraph_edge_index
#>   <int> <int> <LINESTRING>                 <int>
#> 1     1     2   (1 0, 1 2)                     1
#> 2     3     4   (0 1, 2 1)                     2
```

<sup>Created on 2026-07-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

We would need to create a synthetic network  with no possible subdivisions (as the one in the reprex), if we don't want the warning to be raised at all.

Robinlovelace:
So the warning comes from `sf` and is intended behaviour. If this is coming from the tests how about [`expect_warning()`](https://testthat.r-lib.org/reference/expect_error.html) or if it's from the examples you could wrap in a suppress warnings or donttest?

agila5:
Hi @juanfonsecaLS1, I completely understand what you are saying but, actually, the problem is slightly different. The `to_spatial_subdivision` morpher raises a warning whenever it creates a new split and the values are not constant along the splitted geometries. This condition, namely constant values along the geometries, is tested looking at the `agr` of the fields. 

For example: 

``` r
library(sf)
library(tidygraph)
library(sfnetworks)

# Default: value are not constant along geometries
tmp <- as_sfnetwork(roxel["type"]) |> convert(to_spatial_subdivision)
#> Warning: to_spatial_subdivision assumes attributes are constant over geometries

# Define a new data and set 'type' field as constant along geometries
roxel2 <- roxel["type"]
st_agr(roxel2) <- c(type = "constant")
tmp <- as_sfnetwork(roxel2) |> convert(to_spatial_subdivision)
```

<sup>Created on 2026-07-09 with [reprex v2.1.1.9000](https://reprex.tidyverse.org)</sup>

Clearly we cannot enforce such constant agr for every input data so I guess that there is no clear solution. IMO, the easiest strategy is: 1) forget about the warnings in the examples (and fix only if requested during CRAN submission); 2) fix the tests using only toy networks with constant agr attributes, so that the warning is never raised when running the tests. @juanfonsecaLS1 can you work on task 2? Then I will review and merge. 

> So the warning comes from `sf` and is intended behaviour. If this is coming from the tests how about [`expect_warning()`](https://testthat.r-lib.org/reference/expect_error.html) or if it's from the examples you could wrap in a suppress warnings or donttest?

No, the warning is coming from sfnetworks. However, as said above, IMO the easiest way is to just use simpler network that do not create such problem.

juanfonsecaLS1:
Hi both,

I can create a toy network for the tests of `net_2_sfnet_undirected`, but I would need to create a new pbf for testing `oe_get_sfnetwork` as it directly calls oe_get_network. I have tried different `highway_filter` combinations but there is always a possible subdivision that triggers the warning of `sfnetworks`.

As a test, I have implemented the suggestion by @Robinlovelace of using `expect_warning()`. I just pushed a commit with those, @agila5 feel free to revert it if you think it's not correct or it is something CRAN would pick upon

agila5:
Please check the last few commits. If you do not have further comments, I will merge.

juanfonsecaLS1:
This looks good to me. Thanks a lot @agila5

PR #322, opened 2026-08-07
Add oe_get_*network into vignette

### LINKED ISSUE #320: [FEATURE] Process the `oneway` tag

agila5:
Fixes #320 

It's not very detailed but, at least, it is a starting point. Feel free to add more info if you want. 

btw: tomorrow I'm going on holidays, so I will work on a new CRAN release at the end of the month. 

cc @juanfonsecaLS1

### DIFF
--- a/vignettes/tips-and-tricks.Rmd
+++ b/vignettes/tips-and-tricks.Rmd
@@ -59,6 +59,39 @@ out <- oe_get(
 print(out, n = 0L)
 ```
 
+## How can I convert the segments downloaded from OSM into a street network? 
+
+Starting from version 0.7, the package includes a set of functions to automatically convert the output of `oe_get_network()` (and similar functions) into `sfnetwork` or `dodgr` objects. 
+
+For example, the following command returns a `sfnetwork` object representing the walking network extracted from the toy ITS data included in the package: 
+
+```{r}
+sfn_walking = oe_get_sfnetwork(
+  place = "ITS Leeds",
+  mode = "walking", 
+  quiet = TRUE
+)
+```
+
+More precisely, the function runs the following operations: 
+
+  1. It extracts the walkable streets from the toy ITS Leeds data included in the package; 
+  2. It applies a series of preprocessing steps to standardise the values included in the `oneway` column and simplies the `highway` tag removing the `"_link"` suffix. See the `clean_output` argument in `oe_get_network` for more details. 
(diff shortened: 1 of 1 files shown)


Robinlovelace (approved):
Great, let's get this merged! Anything else we can add later.

Issue #323, opened 2026-08-25
Reconstructing way–node incidence from `oe_get()` output — is there a better way?


gustavo-tm:
Hi — first, thanks for the package, it's been really useful.

I'm working on a road safety project where I need the way-node incidence relation.

As far as I can tell this relation isn't available directly from `oe_get()`, since the OSM driver reconstructs geometries and doesn't keep node references. So I've been rebuilding it by exploding the linestrings into vertices and matching those coordinates against the points layer.

The part I'm unsure about is that this only works if I round the coordinates to 7 decimal places before joining — on raw doubles the match fails for a lot of nodes. So in practice I'm snapping and rebuilding the geometries before the join, which feels like something I probably shouldn't need to do. Rounding to 7 dp seems to line up with OSM's native storage precision, so maybe that's the right thing, but I'm not confident.

Is coordinate matching the intended approach here, or is there a supported way to get node references that I've missed?

agila5:
Hi @gustavo-tm! Unfortunately, this is a known limitation of the approach employed in our package and, as far as I know, it's not possible to reconstruct the node/way relationship from GDAL OSM drivers (unless you are using a spatial match operation as you describe).  

Did you already check the data returned by `osmdata`? As far as I know, they are derived from the Overpass API and they preserve the OSM IDs that can be used to match nodes and ways. For example: 

Load packages

``` r
library(sf)
#> Linking to GEOS 3.13.0, GDAL 3.10.1, PROJ 9.5.1; sf_use_s2() is TRUE
library(osmdata)
#> Data (c) OpenStreetMap contributors, ODbL 1.0. https://www.openstreetmap.org/copyright
```

Get OSM roads for a small abandoned town in northern Italy

``` r
cn <- opq("Consonno, Italy", timeout = 75) |> add_osm_feature(key = "highway") |> osmdata_sf()
```

Data are structured as follows.

``` r
cn
#> Object of class 'osmdata' with:
#>                  $bbox : 45.7642766,9.3735149,45.8042766,9.4135149
#>         $overpass_call : The call submitted to the overpass API
#>                  $meta : metadata including timestamp and version numbers
#>            $osm_points : 'sf' Simple Features Collection with 7437 points
#>             $osm_lines : 'sf' Simple Features Collection with 828 linestrings
#>          $osm_polygons : 'sf' Simple Features Collection with 5 polygons
#>        $osm_multilines : NULL
#>     $osm_multipolygons : NULL
```

[...]

PR #325, opened 2026-09-03
Snap coordinates to OSM's native 1e-7 grid on read (#323)

gustavo-tm:
## Description

This adds an internal helper, `snap_to_osm_grid`, in `R/utils.R`, and applies it to the object returned by `my_st_read`. Every layer returned by `oe_get` and `oe_read` therefore has its coordinates rounded onto OSM's native 1e-7 degree grid.

**The problem.** The GDAL OSM driver decodes the same node into slightly different doubles depending on whether it arrives via the `lines` layer or the `points` layer. The difference is in the last bits, far below OSM's own storage resolution, but it is enough to make exact matching fail silently: `st_join` misses points that lie exactly on a line, and vertices exploded with `st_coordinates` do not match the corresponding entries in the points layer. Nothing errors and nothing warns. See #323 for the full diagnosis.

**The approach.** The helper modifies the coordinates rather than setting `st_precision`. `st_precision` only sets an attribute: it is honoured by operations that pass through GEOS or GDAL and ignored by others (notably `st_coordinates`), and it is dropped by `st_transform`. Snapping means every downstream operation sees the same values, and it avoids the sfnetworks
interaction that came up when forcing `st_precision = 1e7` into `oe_read`.

**Escape hatch.** The precision reads from `getOption("osmextract.precision", 1e7)`, so the behaviour can be switched off with `options(osmextract.precision = 0)` without touching any function signature. 

**Scope and limitations.** This is a read-side fix. The `.gpkg` written by `oe_vectortranslate` is unchanged, so a file opened directly with `sf::st_read`, or inspected in QGIS, still carries the unsnapped coordinates. There might be a source-side equivalent, but I'm most confident with the R programming side.

## Related Issue

Relates to #323

## Example

Before, on the current CRAN version:

```r
library(sf)
library(dplyr)
library(osmextract)

lines  <- oe_get("ITS Leeds", layer = "lines",  quiet = TRUE) |> select(osm_id)
points <- oe_get("ITS Leeds", layer = "points", quiet = TRUE) |> select(osm_id)

nrow(lines)   #> 133
nrow(points)  #> 112

joined <- st_join(lines, points)
n_distinct(na.omit(joined$osm_id.y))   #> 6 of 112 points
```

The reason, for one of the missed points:

```r
sprintf("%.17g", c(p[1], v$X))   #> "-1.5525082000000001"  "-1.5525081999999999"
sprintf("%.17g", c(p[2], v$Y))   #> "53.808604800000005"   "53.808604799999998"
```

After, with this branch, the same code gives:

```r
n_distinct(na.omit(joined$osm_id.y))   #> 19
```

The remaining unmatched points are genuinely off the road network (they are nodes tagged with features that do not sit on any imported line), not victims of the precision mismatch.

## Tests

No tests in this PR yet.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -47,6 +47,11 @@ adjust_version_in_url <- function(version, url) {
 # doubles carrying sub-grid noise, which silently breaks exact matching between
 # the layers (see #323). We snap the coordinates back onto the source grid.
 snap_to_osm_grid = function(x, precision = getOption("osmextract.precision", 1e7)) {
+  # Only snap coordinates if the object is an sf object with geometry
+  if (!inherits(x, "sf")) {
+    return(x)
+  }
+  
   sf::st_set_geometry(
     x,
     sf::st_as_sfc(
--- a/R/utils.R
+++ b/R/utils.R
@@ -43,6 +43,17 @@ adjust_version_in_url <- function(version, url) {
   gsub("latest(?=\\.osm\\.pbf$)", version, url, perl = TRUE)
 }
 
+# OSM stores coordinates on a 1e-7 precision grid. The GDAL OSM driver returns
+# doubles carrying sub-grid noise, which silently breaks exact matching between
+# the layers (see #323). We snap the coordinates back onto the source grid.
+snap_to_osm_grid = function(x, precision = getOption("osmextract.precision", 1e7)) {
+  sf::st_set_geometry(
+    x,
+    sf::st_as_sfc(
+      x = sf::st_as_binary(sf::st_geometry(x), precision = precision),
+      crs = sf::st_crs(x))
+  )
+}
 
 # Starting from sf 1.0.2, sf::st_read raises a warning message when both layer
 # and query arguments are set, while it raises a warning in sf < 1.0.2 when
@@ -51,7 +62,7 @@ adjust_version_in_url <- function(version, url) {
 # to circumvent this problem and set the appropriate arguments.
 my_st_read <- function(dsn, layer, quiet, ...) {
   dots_names = ...names()
-  if (utils::packageVersion("sf") <= "1.0.1") { # nocov start
+  out = if (utils::packageVersion("sf") <= "1.0.1") { # nocov start
     sf::st_read(
       dsn = dsn,
       layer = layer,
@@ -74,6 +85,9 @@ my_st_read <- function(dsn, layer, quiet, ...) {
       )
     }
   }
+  
+  # Snap coordinates onto OSM's native 1e-7 precision grid
+  snap_to_osm_grid(out)
 }
 
 #' Returns the download directory used by the package


Robinlovelace:
Good to see tests pass, not sure if this is the best way to implement it but sounds like it works. Maybe worth adding a test or an example that fails without setting precision to OSM's precision. Worth also linking to the OSM wiki that states that latitude and longitude have "7 decimal places": https://wiki.openstreetmap.org/wiki/Node

R/utils.R
```diff
@@ -43,6 +43,22 @@ adjust_version_in_url <- function(version, url) {
   gsub("latest(?=\\.osm\\.pbf$)", version, url, perl = TRUE)
 }
 
+# OSM stores coordinates on a 1e-7 precision grid. The GDAL OSM driver returns
```
Robinlovelace:
```suggestion
# OSM stores coordinates in geographic (lon/lat, EPSG: 4326) coordinates with
# 7 decimal places. The GDAL OSM driver returns
```

agila5:
Hi @gustavo-tm! I'm sorry but I'm not convinced that the suggested approach is the correct one, mainly because `snap_to_osm_grid` does not preserve the information about the `precision` set when constructing the object. In fact, the `st_binary()` call rounds the coordinates to the required precision but, then, does not preserve that information. 

For example: 

``` r
library(sf)
#> Linking to GEOS 3.13.0, GDAL 3.10.1, PROJ 9.5.1; sf_use_s2() is TRUE

nc = st_read(system.file("shape/nc.shp", package="sf"), quiet = TRUE)

snap_to_osm_grid <- function(x, precision = 1e7) {
  sf::st_set_geometry(
    x,
    sf::st_as_sfc(
      x = sf::st_as_binary(sf::st_geometry(x), precision = precision),
      crs = sf::st_crs(x)
    )
  )
}

nc |> snap_to_osm_grid() |> st_precision() # Should be 1e7
#> [1] 0
```

<sup>Created on 2026-09-04 with [reprex v2.1.1.9000](https://reprex.tidyverse.org)</sup>


This is also the reason why you are not facing the error i previously mentioned in the issue related to this PR. 

**EDIT:** I notice that something similar is also mentioned in your post (i.e. the paragraph named **The approach**) but I'm not convinced that silently rounding the coordinates to a given precision and then discarding such transformation is a good idea since it's very easy to forget about this process and then give a false sense of security to other people using these data.

Robinlovelace:
Great reprex @agila5, I was wondering what `st_as_binary` was doing there. Is there a way to set the precision on the GDAL side? Just looking at https://gdal.org/en/stable/development/rfc/rfc99_geometry_coordinate_precision.html ...

gustavo-tm:
@agila5 You're right that the precision attribute isn't preserved, and that was deliberate rather than an oversight. My reasoning: after the coordinates are snapped, the attribute no longer impacts the imported data, it impacts further sf operations that uses input data. In this case, if snapping worked correctly, most operations should not be impacted by setting precision again after snapping. Any operation that will be impacted by precision after snapping are the ones that would have gone off grid and no longer do. Honestly I don't know if there is any function like that and I'm not sure if this should be enforced on the `osmextract` side. Also, just setting precision without snapping is more problematic, since some operation might ignore it and silently break exact joins, such as `st_coordinates`.

It's also worth saying that sf doesn't treat the precision attribute as a record of what was done: `st_transform()` drops it, so `precision = 0` doesn't mean "unrounded". `?st_precision` points at the `st_as_binary` roundtrip as the way to make rounding happen to the R data, which is what I used.


On tests: @Robinlovelace I have one that fails without the change

    points <- oe_get("ITS Leeds", layer = "points", quiet = TRUE)
    mean(round(st_coordinates(points), 7) == st_coordinates(points))
    #> 0.5752688  before
    #> 1          after

agila5:
> You're right that the precision attribute isn't preserved, and that was deliberate rather than an oversight. My reasoning: after the coordinates are snapped, the attribute no longer impacts the imported data, it impacts further sf operations that uses input data. In this case, if snapping worked correctly, most operations should not be impacted by setting precision again after snapping. Any operation that will be impacted by precision after snapping are the ones that would have gone off grid and no longer do. Honestly I don't know if there is any function like that and I'm not sure if this should be enforced on the osmextract side.

I agree with your reasoning but I really don't like the idea of rounding the coordinates to a certain precision level and then completely discard the fact that we ran such operation. Why don't we use `st_precision(out) = 1e7` (or something very similar) in `my_st_read()`? I believe that also fixes your issue.  

> Also, just setting precision without snapping is more problematic, since some operation might ignore it and silently break exact joins, such as st_coordinates.

IMO that's a different problem than what we are discussing here (btw: do you want to open a new issue in sf repo to ask whether st_coordinates should respect a given precision > 0? I can also do that) and a non-issue for your project since, AFAICT, all matching operations are performed by st_join which does not involve st_coordinates and (should) take precision into account. 

> It's also worth saying that sf doesn't treat the precision attribute as a record of what was done: st_transform() drops it, so precision = 0 doesn't mean "unrounded".

I'm not a geographer, but I believe that st_transform drops a given precision since we cannot easily convert lat/long precision into easting/northing precision (https://wiki.openstreetmap.org/wiki/Precision_of_coordinates). In any case, I believe this is also a different issue than what we are discussing here.

gustavo-tm:
> I really don't like the idea of rounding the coordinates to a certain precision level and then completely discard the fact that we ran such operation

This is fair. I think that pointing this out on the `oe_read` and `oe_get` or on a vignette is important. However, if we can agree that any decimal places after the seventh are mistakes, we should agree that some sort of snapping procedure is necessary.

> Why don't we use st_precision(out) = 1e7 (or something very similar) in my_st_read()? 

I don't have a problem with setting precision after reading in `my_st_read`. In fact, I quite like the idea. I think it gives  transparency about the fact that the native grid resolution is 1e7 and the rounding is intentional and important. Any worried user that sees the precision setting on their dataframe on import should easily find an explanation on the documentation. 

I just proposed snap instead of using st_precision to avoid the downstream problems you mentioned about networks on #323. If precision is set after snapping, then it should be a no-op, but would add to transparency. 

If no snapping is done before, there can be issues. I think it would be important to make sure that all sf commands use it and the tag persists. For instance, joining points and lines when both have the same precision keeps the tag. On the other hand, if joining with geometries that don't have a precision tag, after the operation the tag is lost and the points are not snapped. This can be a problem if the user joins the precision points with their own geometries before joining with lines, as in this case the tag will be lost and the join will fail silently.

``` r
library(sf)
library(osmextract)

lines  <- oe_get("ITS Leeds", layer = "lines",  quiet = TRUE) 
points <- oe_get("ITS Leeds", layer = "points", quiet = TRUE)

st_join(st_set_precision(lines, 1e7), st_set_precision(points, 1e7)) |> st_precision() #> 1e+07
st_join(lines, st_set_precision(points, 1e7)) |> st_precision() #> 0
``` 

> I believe that also fixes your issue.

It does in my case because right after import I immediately use `st_transform` which does the rounding before the transformation, effectively snapping to the grid

> IMO that's a different problem than what we are discussing here (btw: do you want to open a new issue in sf repo to ask whether st_coordinates should respect a given precision > 0? I can also do that) and a non-issue for your project since, AFAICT, all matching operations are performed by st_join which does not involve st_coordinates and (should) take precision into account.

This could be the case, yes. Maybe someone from sf could give a richer perspective about what commands take precision into account and if this behavior in `st_coordinates` is in fact intended. To be honest, I think that whoever created this option in sf did not have this use case in mind. See the comment on their documentation:

"The reason to support precision is that geometrical operations in GEOS or liblwgeom may work better at reduced precision. For writing data from R to external resources it is harder to think of a good reason to limiting precision."

I opened an issue on sf https://github.com/r-spatial/sf/issues/2620

agila5:
Hi @gustavo-tm! 

> This is fair. I think that pointing this out on the oe_read and oe_get or on a vignette is important. However, if we can agree that any decimal places after the seventh are mistakes, we should agree that some sort of snapping procedure is necessary.

I started implementing this idea in #326. I will finish the PR soon but, in any case, we need dev-sf to be released on CRAN before we can also release new approach to CRAN. 

> I just proposed snap instead of using st_precision to avoid the downstream problems you mentioned about networks on https://github.com/ropensci/osmextract/issues/323. If precision is set after snapping, then it should be a no-op, but would add to transparency.

This is now fixed in https://github.com/r-spatial/sf/pull/2619 but, unfortunately, not available on CRAN yet

> Maybe someone from sf could give a richer perspective about what commands take precision into account and if this behavior in st_coordinates is in fact intended. 

This is now partially implemented in https://github.com/r-spatial/sf/pull/2622 (+ additional changes in the main branch) and available on CRAN!

Robinlovelace:
Just to add: we can temporarily merge to main with 

```
Remotes: r-spatial/sf
```

before we submit to CRAN while we await the next CRAN version of {sf}, could also be good to merge to main and ask others to test it to help find any issues.

Robinlovelace:
> merge to main

Merge https://github.com/ropensci/osmextract/pull/326 to main I mean.

gustavo-tm:
Hey, @agila5. Cool to see the implementation on sf. 

However, this point still stands

> If no snapping is done before, there can be issues. [...] For instance, joining points and lines when both have the same precision keeps the tag. On the other hand, if joining with geometries that don't have a precision tag, after the operation the tag is lost and the points are not snapped. This can be a problem if the user joins the precision points with their own geometries before joining with lines, as in this case the tag will be lost and the join will fail silently.

Reprex
``` r
library(sf)
library(osmextract)

lines  <- oe_get("ITS Leeds", layer = "lines",  quiet = TRUE) 
points <- oe_get("ITS Leeds", layer = "points", quiet = TRUE)

st_join(st_set_precision(lines, 1e7), st_set_precision(points, 1e7)) |> st_precision() #> 1e+07
st_join(lines, st_set_precision(points, 1e7)) |> st_precision() #> 0
```

This is why I still believe that a binary roundtrip for coordinate snapping, as suggested in the sf package documentation and implemented in this PR is the correct call here. We can add precision too after the snapping for transparency. I will add a commit with an implementation for that in this PR.

However, feel free to close if you believe this not relevant.

Robinlovelace:
This makes a strong case for snapping. So this PR and the subsequent one from @agila5 in #326 are the 2 options for tackling the bug, right? 

A few high-level questions:

- Are the 2 PRs mutually exclusive?
- What are pros and cons of each?

Hope that's useful general thoughts, I don't have a strong view on this but I don think we should tackle the underlying issue which I think we're doing. I also think the final decision should be one that Andrea is happy with as lead developer of this package so keen to hear from him about pros and cons.

agila5:
Hi both! 

@gustavo-tm: I'm sorry but I don't understand the problem within your example. Could you please explain it a bit better? I guess the issue is that `st_precision` returns different results in the two cases. However, the underlying cause is that `st_join` preserves the precision of the first input (which is a reasonable behaviour IMO) and, in your example, the object `lines` was built with 0 precision (which is the default behaviour, for example, in the current CRAN version of the package). 

However, after #326 is merged, you would get identical results in both cases: 

``` r
remotes::install_github("ropensci/osmextract#326")
library(sf)
library(osmextract)

lines  <- oe_get("ITS Leeds", layer = "lines",  quiet = TRUE) 
points <- oe_get("ITS Leeds", layer = "points", quiet = TRUE)

st_join(st_set_precision(lines, 1e7), st_set_precision(points, 1e7)) |> st_precision()
#> [1] 1e+07
st_join(lines, st_set_precision(points, 1e7)) |> st_precision()
#> [1] 1e+07
```

<sup>Created on 2026-09-17 with [reprex v2.1.1.9000](https://reprex.tidyverse.org)</sup>

@Robinlovelace: 

> Are the 2 PRs mutually exclusive? 

I guess so

> What are pros and cons of each?

I prefer #326 since it achieves the same results as the current PR (apart from existing sf functions that might ignore the precision arg, such as the st_coordinates case) with minimal effort (we just set an attribute and nothing else) and without rebuilding the geometries every time we read a new OSM extract. I agree that there might be frictions between the proposed approach and some sf functions, but IMO such frictions should be solved in sf (as we did with st_coordinates) or patched by a user (in case the friction cannot be smoothed out).

Robinlovelace:
👍 to @agila5's approach. If you can get back to us with any further thoughts and/or reproducible examples showing unexpected results on that branch let us know @gustavo-tm 

You can checkout Andrea's branch with something like the following after installing the gh CLI tool:

```
gh repo clone ropensci/osmextract
cd osmextract
gh pr checkout 326
# Then in R: devtools::install() 
```

Issue #327, opened 2026-09-11
[BUG] net_2_sfnet_undirected() fails with dev sfnetworks


agila5:
**Describe the bug**
The approach we currently use in `net_2_sfnet_undirected()` does not work with `sfnetwork` v1.0 (which is currently under development). Hence, when sfnetwork code is stable, we need to update it and, possibly, support both versions. 

**Reproducible example**
Just test the package with github version of sfnetwork to see the problem.

juanfonsecaLS1:
Hi @agila5, I just tested this and came up with this implementation for the `net_2_sfnet_undirected` which should work with both versions of `sfnetworks`. 

Let me know what you think, I can create a PR with this and updating the tests.

Thanks again for all your work!

```r
net_2_sfnet_undirected = function(net_sf, require_equal = TRUE) {

  if (!requireNamespace("sfnetworks", quietly = TRUE)) {
    stop("sfnetworks is not available. Please install it first")
  }

  stopifnot(
    (is.logical(require_equal) && length(require_equal) == 1L) ||
      (is.character(require_equal) && all(require_equal %in% names(net_sf)))
  )

  sfnet = sfnetworks::as_sfnetwork(
    x = net_sf,
    directed = FALSE
  )

  # Creating junctions where road segments overlap
  # This converts implicit intersections into explicit nodes
  sf_net_subdiv = tidygraph::convert(sfnet, sfnetworks::to_spatial_subdivision, .clean = TRUE)

# This converts the require_equal argument into a character vector if TRUE (Works with both implementations of sfnetworks) 
if(is.logical(require_equal)&&require_equal){
    require_equal = names(sf::st_drop_geometry(net_sf))
  } 


if(packageVersion("sfnetworks")<package_version("0.9.9.9000")){
# Implementation working with current version of sfnetworks on CRAN
# Simplifying the interstitial nodes segments 
  tidygraph::convert(
    sf_net_subdiv,
    sfnetworks::to_spatial_smooth,
    summarise_attributes = list(collapse_function),
    require_equal = require_equal,
    .clean = TRUE
  )

} else {
  # if require_equal is FALSE, it is set to NULL as per new implementation in sfnetworks>0.9
  if(is.logical(require_equal)&&!require_equal){
    require_equal = NULL
  }
  
  tidygraph::convert(
      sf_net_subdiv,
      sfnetworks::to_spatial_smooth,
      attribute_summary = list(collapse_function),
      require_equal = require_equal,
[...]

PR #326, opened 2026-09-11
Add precision to output

agila5:
Forcefully set `precision = 1e7` to the output of oe_read (and, as a consequence, oe_get(), oe_get_network() and related functions). 

## Description
See above. 

## Related Issue
See discussion in #323 and #325. 

## Example
``` r
library(osmextract)
#> Data (c) OpenStreetMap contributors, ODbL 1.0. https://www.openstreetmap.org/copyright.
#> Check the package website, https://docs.ropensci.org/osmextract/, for more details.
its <- oe_get("ITS Leeds", quiet = TRUE)
sf::st_precision(its)
#> [1] 1e+07
```

<sup>Created on 2026-09-16 with [reprex v2.1.1.9000](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.4.2 (2024-10-31 ucrt)
#>  os       Windows 11 x64 (build 26200)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  Italian_Italy.utf8
#>  ctype    Italian_Italy.utf8
#>  tz       Europe/Rome
#>  date     2026-09-16
#>  pandoc   3.2 @ C:/Program Files/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)
#>  quarto   1.6.42 @ C:\\Users\\andre\\AppData\\Local\\Programs\\Quarto\\bin\\quarto.exe
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date (UTC) lib source
#>  class         7.3-22     2023-05-03 [2] CRAN (R 4.4.2)
#>  classInt      0.4-11     2025-01-08 [1] CRAN (R 4.4.2)
#>  cli           3.6.6      2026-04-09 [1] CRAN (R 4.4.2)
#>  DBI           1.3.0      2026-02-25 [1] CRAN (R 4.4.2)
#>  digest        0.6.37     2024-08-19 [1] CRAN (R 4.4.2)
#>  e1071         1.7-17     2025-12-18 [1] CRAN (R 4.4.2)
#>  evaluate      1.0.5      2025-08-27 [1] CRAN (R 4.4.2)
#>  fastmap       1.2.0      2024-05-15 [1] CRAN (R 4.4.2)
[...]

PR #328, opened 2026-09-17
update implementation of `net_2_sfnet_undirected` to work with sfnetworks v1.0

### LINKED ISSUE #4: Download based on a geographic input?

juanfonsecaLS1:
## Description
Improves the implementation of `net_2_sfnet_undirected()` by checking the installed version of `sfnetworks` and calling the `to_spatial_smooth` with the corresponding parameters. 

## Related Issue

Addresses #327

## Example

It does not change the behaviour of the functions

Tests did not need to be updated as behaviour is identical. I successfully ran the tests in two environments with the relevant versions of `sfnetworks`: 0.6.6 (current CRAN) and 0.9.9.9000 (GitHub dev version)


Robinlovelace:
Looks good to me @juanfonsecaLS1, many thanks! Out of interest, is this affected by planned upstream changes to set 10^-7 as the resolution of OSM data, defined in the OSM wiki and implemented by @agila5 in #326 ?

I imagine not but just heads-up on parallel work going on.

juanfonsecaLS1:
It should not be related. So far, the only changes in this PR are link to a name change of one of the arguments taken by `to_spatial_smooth`, and the values the `require_equal` take. 

I will also push a small change to improve the error message thrown when `require_equal` does not meet the requirements.

Question for both of you @Robinlovelace and @agila5. At the moment, `require_equal` takes `TRUE/FALSE` or a character vector with column names, as per the current CRAN version of `sfnetworks`. Should we make it simpler, as in the new implementation, only character?

Robinlovelace:
> Question for both of you @Robinlovelace and @agila5. At the moment, `require_equal` takes `TRUE/FALSE` or a character vector with column names, as per the current CRAN version of `sfnetworks`. Should we make it simpler, as in the new implementation, only character?

I don't currently have a preference and not sure of the prose/cons. Simple can be good but can also see benefit if `TRUE`. Interested to learn your thoughts on this Juan, and Andrea's.

agila5:
> Should we make it simpler, as in the new implementation, only character?

IMO no because the current implementation of dev sfnetwork is not stable and might change in the future. I agree that we can and we should support it, but I wouldn't change the existing implementation (which is based on a working package on CRAN) until sfnetwork 1.0 is in its final state.

R/get-network-graph.R
```diff
@@ -219,14 +220,37 @@ net_2_sfnet_undirected = function(net_sf, require_equal = TRUE) {
   # This converts implicit intersections into explicit nodes
   sf_net_subdiv = tidygraph::convert(sfnet, sfnetworks::to_spatial_subdivision, .clean = TRUE)
 
-  # Simplifying the interstitial nodes segments
+# This converts the require_equal argument into a character vector if TRUE (Works with both implementations of sfnetworks) 
+if(is.logical(require_equal)&&require_equal){
+    require_equal = names(sf::st_drop_geometry(net_sf))
+  } 
+
+if(utils::packageVersion("sfnetworks")<"0.9"){
```
agila5: Why "0.9"? I guess that the new dev version will be release as 1.0.

R/get-network-graph.R
```diff
@@ -219,14 +220,37 @@ net_2_sfnet_undirected = function(net_sf, require_equal = TRUE) {
   # This converts implicit intersections into explicit nodes
   sf_net_subdiv = tidygraph::convert(sfnet, sfnetworks::to_spatial_subdivision, .clean = TRUE)
 
-  # Simplifying the interstitial nodes segments
+# This converts the require_equal argument into a character vector if TRUE (Works with both implementations of sfnetworks) 
+if(is.logical(require_equal)&&require_equal){
```
agila5: You can achieve the same result with less code using `if (isTRUE(require_equal)) {...}`

R/get-network-graph.R
```diff
@@ -219,14 +220,37 @@ net_2_sfnet_undirected = function(net_sf, require_equal = TRUE) {
...
+    require_equal = names(sf::st_drop_geometry(net_sf))
```
agila5: I'm not sure I understand why this change is required. Can you please explain a little bit better?

R/get-network-graph.R
```diff
@@ -219,14 +220,37 @@ net_2_sfnet_undirected = function(net_sf, require_equal = TRUE) {
...
     sf_net_subdiv,
     sfnetworks::to_spatial_smooth,
     summarise_attributes = list(collapse_function),
     require_equal = require_equal,
     .clean = TRUE
   )
+
+} else {
+  # if require_equal is FALSE, it is set to NULL as per new implementation in sfnetworks>0.9
+  if(is.logical(require_equal)&&!require_equal){
```
agila5: You can achieve the same result (and with less code) using `if (isFALSE(require_equal)) {...}`

juanfonsecaLS1:
Since modifying the current implementation wouldn't bring any benefits, and, as @agila5 has mentioned, dev `sfnetworks` might change, I will stick to the current implementation. I just pushed some changes with a clearer error message for invalid `require_equal` and the corresponding tests.

agila5:
Hi @juanfonsecaLS1! Ok, thanks. Please adjust the reference sfnetworks version (0.9 --> 1.0) and, if you don't mind, remove the tests + additional `names(sf::st_drop_geometry(...))` step (or please explain why it's important). Then I will merge!

juanfonsecaLS1:
Hi @agila5 , thanks for this!

> Please adjust the reference sfnetworks version (0.9 --> 1.0)

I initially used 1.0; however that fails when using the current development version (0.9.9.9000), should I make it `>=0.9.9.9000`?


> remove the tests + additional names(sf::st_drop_geometry(...)) step (or please explain why it's important)

I have removed the redundant test. For the additional step, I have improved the implementation avoiding `st_drop_geometry()`. Dropping the geometry column is needed because `to_spatial_smooth` would not simplify/merge
adjacent links at all. For example:

``` r
collapse_function = function(x) {
  paste(unique(x), collapse = ",")
}

net_sf = sf::st_sf(
    highway = c("residential", "residential"),
    attr_a = c(123, 123),
    attr_b = c("asdasd", "asdasd"),
    geometry = sf::st_sfc(
      sf::st_linestring(rbind(c(0, 0), c(1, 0))),
      sf::st_linestring(rbind(c(1, 0), c(2, 0))),
      crs = 4326
    )
  )

sfnet = sfnetworks::as_sfnetwork(
    x = net_sf,
    directed = FALSE
  )


geom_col <- attr(net_sf, "sf_column")
    

tidygraph::convert(
      sfnet,
      sfnetworks::to_spatial_smooth,
      summarise_attributes = list(collapse_function),
      require_equal = names(net_sf),
      .clean = TRUE
    )
#> # A sfnetwork with 3 nodes and 2 edges
#> #
#> # CRS:  EPSG:4326 
#> #
#> # An unrooted tree with spatially explicit edges
#> #
#> # Node data: 3 × 1 (active)
#>      geometry
#>   <POINT [°]>
#> 1       (0 0)
#> 2       (1 0)
#> 3       (2 0)
#> #
#> # Edge data: 2 × 6
#>    from    to highway     attr_a attr_b         geometry
#>   <int> <int> <chr>        <dbl> <chr>  <LINESTRING [°]>
#> 1     1     2 residential    123 asdasd       (0 0, 1 0)
#> 2     2     3 residential    123 asdasd       (1 0, 2 0)


tidygraph::convert(
      sfnet,
      sfnetworks::to_spatial_smooth,
      summarise_attributes = list(collapse_function),
      require_equal = setdiff(names(net_sf), geom_col),
      .clean = TRUE
    )
#> # A sfnetwork with 2 nodes and 1 edges
#> #
#> # CRS:  EPSG:4326 
#> #
#> # An unrooted tree with spatially explicit edges
#> #
#> # Node data: 2 × 1 (active)
#>      geometry
#>   <POINT [°]>
#> 1       (0 0)
#> 2       (2 0)
#> #
#> # Edge data: 1 × 6
#>    from    to highway     attr_a attr_b         geometry
#>   <int> <int> <chr>       <chr>  <chr>  <LINESTRING [°]>
#> 1     1     2 residential 123    asdasd  (0 0, 1 0, 2 0)
```

<sup>Created on 2026-09-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

agila5:
Ok, it's clear. Thanks! Gonna merge.

